PR #18, opened 2021-12-08
Upgrade Elasticsearch to 6.8.20

jeffw16:


jeffw16:
Tested and it works. Merging

PR #19, opened 2021-12-09
Add support for Caddy TLS termination proxy

jeffw16:
This moves the Canasta app image away from providing any ports to the host. Caddy, the TLS termination proxy, will instead serve to the host ports 80 and 443.


jeffw16:
By the way, I think there's some issues with using non-port 80 ports on the Canasta image. This PR incorporates a change to `ports.conf`, but it doesn't seem to fully resolve the problem.

jeffw16:
I just tried it out and it seems to be working!

jeffw16:
And here is an example of Caddy successfully automatically provisioning a TLS certificate:

```
ubuntu@ip-172-31-25-239:~/Canasta$ sudo docker logs canasta_caddy_1
Caddy proxying https://canastatest1.mywikis.net -> http://web:8080
{"level":"warn","ts":1643367431.557757,"logger":"admin","msg":"admin endpoint disabled"}
{"level":"info","ts":1643367431.557922,"logger":"http","msg":"server is listening only on the HTTPS port but has no TLS connection policies; adding one to enable TLS","server_name":"proxy","https_port":443}
{"level":"info","ts":1643367431.5579395,"logger":"http","msg":"enabling automatic HTTP->HTTPS redirects","server_name":"proxy"}
{"level":"info","ts":1643367431.558107,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc00019fd50"}
{"level":"info","ts":1643367431.558259,"logger":"tls","msg":"cleaning storage unit","description":"FileStorage:/data/caddy"}
{"level":"info","ts":1643367431.5584235,"logger":"http","msg":"enabling automatic TLS certificate management","domains":["canastatest1.mywikis.net"]}
{"level":"info","ts":1643367431.5586016,"msg":"autosaved config (load with --resume flag)","file":"/config/caddy/autosave.json"}
{"level":"info","ts":1643367431.5586364,"logger":"tls","msg":"finished cleaning storage units"}
{"level":"info","ts":1643367431.5589042,"logger":"tls.obtain","msg":"acquiring lock","identifier":"canastatest1.mywikis.net"}
{"level":"info","ts":1643367431.5676818,"logger":"tls.obtain","msg":"lock acquired","identifier":"canastatest1.mywikis.net"}
{"level":"info","ts":1643367431.5889199,"logger":"tls.issuance.acme","msg":"waiting on internal rate limiter","identifiers":["canastatest1.mywikis.net"],"ca":"https://acme-v02.api.letsencrypt.org/directory","account":""}
{"level":"info","ts":1643367431.5889456,"logger":"tls.issuance.acme","msg":"done waiting on internal rate limiter","identifiers":["canastatest1.mywikis.net"],"ca":"https://acme-v02.api.letsencrypt.org/directory","account":""}
{"level":"info","ts":1643367432.0219877,"logger":"tls.issuance.acme.acme_client","msg":"trying to solve challenge","identifier":"canastatest1.mywikis.net","challenge_type":"http-01","ca":"https://acme-v02.api.letsencrypt.org/directory"}
{"level":"error","ts":1643367432.3920736,"logger":"tls.issuance.acme.acme_client","msg":"challenge failed","identifier":"canastatest1.mywikis.net","challenge_type":"http-01","problem":{"type":"urn:ietf:params:acme:error:connection","title":"","detail":"Fetching http://canastatest1.mywikis.net/.well-known/acme-challenge/eprvmH37mhrig3hiFiz6mf8W_PbCSbRne2n7VOT6avk: Connection refused","instance":"","subproblems":[]}}
{"level":"error","ts":1643367432.392186,"logger":"tls.issuance.acme.acme_client","msg":"validating authorization","identifier":"canastatest1.mywikis.net","problem":{"type":"urn:ietf:params:acme:error:connection","title":"","detail":"Fetching http://canastatest1.mywikis.net/.well-known/acme-challenge/eprvmH37mhrig3hiFiz6mf8W_PbCSbRne2n7VOT6avk: Connection refused","instance":"","subproblems":[]},"order":"https://acme-v02.api.letsencrypt.org/acme/order/382719210/59162756120","attempt":1,"max_attempts":3}
{"level":"info","ts":1643367433.5670924,"logger":"tls.issuance.acme.acme_client","msg":"trying to solve challenge","identifier":"canastatest1.mywikis.net","challenge_type":"tls-alpn-01","ca":"https://acme-v02.api.letsencrypt.org/directory"}
{"level":"info","ts":1643367433.7978704,"logger":"tls","msg":"served key authentication certificate","server_name":"canastatest1.mywikis.net","challenge":"tls-alpn-01","remote":"52.39.4.59:51558","distributed":false}
{"level":"info","ts":1643367433.8218176,"logger":"tls","msg":"served key authentication certificate","server_name":"canastatest1.mywikis.net","challenge":"tls-alpn-01","remote":"3.19.56.43:23134","distributed":false}
{"level":"info","ts":1643367433.897994,"logger":"tls","msg":"served key authentication certificate","server_name":"canastatest1.mywikis.net","challenge":"tls-alpn-01","remote":"66.133.109.36:55564","distributed":false}
{"level":"info","ts":1643367433.990173,"logger":"tls","msg":"served key authentication certificate","server_name":"canastatest1.mywikis.net","challenge":"tls-alpn-01","remote":"18.192.36.99:50506","distributed":false}
{"level":"info","ts":1643367434.251371,"logger":"tls.issuance.acme.acme_client","msg":"validations succeeded; finalizing order","order":"https://acme-v02.api.letsencrypt.org/acme/order/382719210/59162761660"}
{"level":"info","ts":1643367434.6734843,"logger":"tls.issuance.acme.acme_client","msg":"successfully downloaded available certificate chains","count":2,"first_url":"https://acme-v02.api.letsencrypt.org/acme/cert/039f644429de4d0b280d335d73fa11181054"}
{"level":"info","ts":1643367434.6751308,"logger":"tls.obtain","msg":"certificate obtained successfully","identifier":"canastatest1.mywikis.net"}
{"level":"info","ts":1643367434.675274,"logger":"tls.obtain","msg":"releasing lock","identifier":"canastatest1.mywikis.net"}
```

docker-compose.yml
```diff
@@ -53,6 +51,16 @@ services:
...
 
+  caddy:
+    image: caddy:2.4.6-alpine
+    restart: unless-stopped
+    entrypoint: caddy reverse-proxy --from ${MW_SITE_FQDN:-localhost} --to web:8080
+    ports:
+      - "${PORT:-80}:80"
+      - "443:443"
+    volumes:
+      - ./_caddy_data/:/data
```
vedmaka, line 62: I know the data-dir need to be persistent since it stores certificates and stuff there, though since on the Canasta we do store everything on volumes by default (even DB data, see `mysql-data-volume` at the bottom) I'd also propose to store caddy data-dir as a docker volume instead of mounting a directory into a container

_sources/configs/ports.conf
```diff
@@ -0,0 +1,15 @@
+# If you just change the port or add more ports here, you will likely also
+# have to change the VirtualHost statement in
+# /etc/apache2/sites-enabled/000-default.conf
+
+Listen 8080
```
vedmaka, line 5: I think we should also update the `EXPOSE` instruction here https://github.com/WikiWorks/Canasta/blob/master/Dockerfile#L667

docker-compose.yml
```diff
@@ -53,6 +51,16 @@ services:
     volumes:
       - elasticsearch:/usr/share/elasticsearch/data
 
+  caddy:
+    image: caddy:2.4.6-alpine
+    restart: unless-stopped
+    entrypoint: caddy reverse-proxy --from ${MW_SITE_FQDN:-localhost} --to web:8080
+    ports:
+      - "${PORT:-80}:80"
+      - "443:443"
```
vedmaka, line 60: This actually could be a problem when there is a need to spin up two wikis on the same machine, eg: production and test environments.

With external Nginx acting as TLS terminator this was solved easily by having two stacks deployed separately with different ports set for each deployment via `.env` and Nginx configured to route traffic to specific ports matching via domain.

With this new setup when Caddy is handling TLS termination, the approach for deploying multiple environments requires a change - apparently, the Caddy is quite flexible and can serve as a full replacement for external Nginx [1] thus the solution could be to store all the environments under the same `docker-compose` file and provide Caddy with a config to declare routes, etc

Storing all the environments in a single compose file definitely has its own drawbacks though, but I can't think of a better solution, please share your thoughts

1. https://caddyserver.com/docs/caddyfile/patterns

docker-compose.yml
jeffw16: I agree completely

docker-compose.yml
jeffw16: I agree. In my opinion:

- For anyone using Canasta, using Caddy is optional and can be completely removed from the Docker Compose manifest if they want to use Nginx instead
- If they really want to use Caddy for a multi-wiki setup, then they can modify the Docker Compose manifest to host multiple wikis, as you said

This addition of Caddy is just an option for people; it doesn't have to be a required component of Canasta, imo.

vedmaka: Ah, this makes good sense, for some reason I did not treat this as an optional thing, though I think in that case I'd propose to still try to keep the Apache running on port `80` on `ports.conf` for backward compatibility, you did mention above there were issues doing that, what kind of issues did you experience?

jeffw16: @vedmaka Caddy likes to have port 80 available so Let's Encrypt can perform verification and also it can enforce 80 -> 443 redirection. That being said, I think this can be disabled and perhaps we can configure Caddy to use another port. https://caddyserver.com/docs/json/apps/http/servers/automatic_https/

_sources/configs/ports.conf
jeffw16: Hmm, I wonder why it is working right now without `EXPOSE 8080`?

docker-compose.yml
jeffw16: Actually, ACME requires the challenge to be performed on port 80. There is not a way for us to change that. Since the certificate could need to be renewed at any time, Caddy requires the ability to control port 80, or at least have access to port 80 somehow so it can do ACME-related challenges, etc. https://letsencrypt.org/docs/challenge-types/

Perhaps it will be possible to run Apache on port 80, but all port conflicts must be avoided.

vedmaka: @jeffw16 yep, sorry for confusing you, I did mean to run the `web` container on port `80` to keep it backward compatible with older image versions, my idea was that `caddy reverse-proxy --from ${MW_SITE_FQDN:-localhost} --to web:80` should work out of the box, considering the `web` port won't be exposed to the host and thus Caddy will be able to use and expose `80` and at the same time communicate with the `web` container on port `80` too within the docker network

I did not try that though, do you think this setup could cause any conflicts?

jeffw16: I think it does cause conflicts. I just tried doing what you said and did `expose: "80"` for the `web` container. Unfortunately I think Apache is listening on `*` or`127.0.0.1` and this is probably why it is causing some sort of conflict.

jeffw16: Here is what I am getting when I try to do `curl` from within the Canasta container:

```
root@5869bbcf9a57:/var/www/mediawiki/w# curl web:80
curl: (7) Failed to connect to web port 80: Connection refused
```

This issue doesn't happen when the Canasta port is changed to 8080.

vedmaka: @jeffw16 I've just tried to reproduce this and was not able to do so, on my side it worked well:

1. I've cloned the PR
2. modified Caddy `entrypoint` to `caddy reverse-proxy --from ${MW_SITE_FQDN:-localhost} --to web:80`
3. ensured the `web` service definition **does not** expose any port
4. ssh'ed into `caddy` container and ran `wget http://web:80`, got:

```
Connecting to web:80 (172.18.0.5:80)
wget: server returned error: HTTP/1.1 500 Internal Server Error
```

and here is from the `web` container:

```
root@21369f4aaa9c:/var/www/mediawiki/w# curl -I -XGET web:80
HTTP/1.1 500 Internal Server Error
Date: Wed, 02 Feb 2022 10:55:50 GMT
Server: Apache/2.4.38 (Debian)
X-Content-Type-Options: nosniff
Content-Length: 435
X-Request-Id: 5ad300dfe3869e017e8a3c84
Connection: close
Content-Type: text/html; charset=utf-8
```

(ignore 500 internal error, that's because I did not have DB initialized)

this indicates that `caddy` is able to communicate with the `web` just fine via port `80`

> I just tried doing what you said and did expose: "80" for the web container

I think that's the problem, we should not expose any port from the `web` container to the host and instead only expose `caddy` port to a host

jeffw16: Ok, I managed to get it working when I cloned a fresh copy of the repo. That's really weird, I don't know why that happened. I will make the changes to the repo now to fix up everything.

Btw, about `ports` vs `expose`: `expose` is supposed to expose ports between containers but doesn't expose it to the host; that's what `ports` is for. Actually, it looks like `expose` doesn't even do anything anymore; it's just there for reference. https://stackoverflow.com/questions/40801772/what-is-the-difference-between-docker-compose-ports-vs-expose

sanjay-thiyagarajan:
@jeffw16 I've tried to clone the branch and run ```docker-compose up``` . It forces the port to be 443 (https). And in that case, I'm not taken to wiki-creation wizard when LocalSettings.php is not found.

jeffw16:
@sanjay-thiyagarajan Looks like there is a step that I forgot to do: specify to Caddy to use a different port.

I don't see a straightforward way to do it using the `caddy reverse-proxy` command, so it's likely Canasta will have to switch to using a Caddyfile. I don't see an easy way to do it right now, proposals are welcome. https://caddyserver.com/docs/caddyfile/options

In the meantime, I would say let's just merge it without support for non-standard HTTPS port numbers. Worst case, the user can disable Caddy and restore it to using Apache.

sanjay-thiyagarajan:
So, 

> @sanjay-thiyagarajan Looks like there is a step that I forgot to do: specify to Caddy to use a different port.
> 
> I don't see a straightforward way to do it using the `caddy reverse-proxy` command, so it's likely Canasta will have to switch to using a Caddyfile. I don't see an easy way to do it right now, proposals are welcome. https://caddyserver.com/docs/caddyfile/options
> 
> In the meantime, I would say let's just merge it without support for non-standard HTTPS port numbers. Worst case, the user can disable Caddy and restore it to using Apache.

So, how do I get to https://localhost ? Doing ```docker-compose up``` was not taking me to my canasta wiki / setup wizard

sanjay-thiyagarajan:
Okay, In that case, everything works fine ! All traffic to http gets redirected to https now.

yaronkoren:
Yes, it seems like this should just be merged in... should I do it?

jeffw16:
Sure, go for it!

PR #27, opened 2022-01-28
Remove the security-related globals

tosfos:
These were added to address CVE-2021-44858, CVE-2021-45038, CVE-2021-44857, but they are not needed any more, since the image uses 1.35.5.


jeffw16 (approved):
LGTM

PR #34, opened 2022-02-15
Fix for loading of MW skins (eliminates the need for symlinks)

sanjay-thiyagarajan:


jeffw16:
This is a good idea. What about also setting [`$wgStyleDirectory` (see the MW.o page)](https://www.mediawiki.org/wiki/Manual:$wgStyleDirectory), could we do that too?

jeffw16:
Also, @vedmaka, I believe there is an issue with the GitHub Actions publish stage when an external branch is used. Can you take a look at this when you have time?

sanjay-thiyagarajan:
@jeffw16 $wgStyleDirectory is being set to the right value in cfLoadSkin function. But in the case of wfStylePath, it was not at all being set.

PR #36, opened 2022-02-26
Feature/sitemap ping

vedmaka:


_sources/scripts/mwsitemapgen.sh
```diff
@@ -21,6 +21,10 @@ while true; do
       --skip-redirects \
       --identifier=$MW_SITEMAP_IDENTIFIER
 
+    # sending the sitemap to google
+    echo "sending to Google -> https://www.google.com/ping?sitemap=$MW_SITE_SERVER/$MW_SCRIPT_PATH/sitemap/$MW_SITEMAP_SUBDIR/sitemap-index-$MW_SITEMAP_IDENTIFIER.xml"
```
pastakhov, line 25: I think we don't need / before $WG_SCRIPT_PATH because it already starts with /. (slash is doubled now)
slash is doubled also when MW_SITEMAP_SUBDIR is empty.
Also $WG_SITE_SERVER can be a protocol-relative url like `//site.name`

pastakhov (changes requested):

_sources/scripts/run-apache.sh
```diff
@@ -147,11 +147,11 @@ sitemapgen() {
         fi
         # Fall back to default value if can't fetch the variable
         if [ -z "$MW_SCRIPT_PATH" ]; then
-          MW_SCRIPT_PATH="w"
+          MW_SCRIPT_PATH="/w"
         fi
         export MW_SCRIPT_PATH
         echo >&2 Run sitemap generator
-        nice -n 20 runuser -c /mwsitemapgen.sh -s /bin/bash "$WWW_USER"
+        MW_SCRIPT_PATH=$MW_SCRIPT_PATH nice -n 20 runuser -c /mwsitemapgen.sh -s /bin/bash "$WWW_USER"
```
pastakhov, line 154: It looks like it was added by mistake, is not?

pastakhov (changes requested):

_sources/scripts/run-apache.sh
vedmaka: nope, actually the line 152 not removed is a mistake ;)

amalpaul54111:
Does this issue #28 still exist in your current version?

_sources/scripts/run-apache.sh
pastakhov: ok then

pastakhov (approved):

PR #47, opened 2022-03-11
Reorganize LocalSettings and composer

pastakhov:


jeffw16:
Although this is a draft, I think it looks good so far!

_sources/canasta/CanastaDefaultSettings.php
```diff
@@ -0,0 +1,73 @@
...
+
+function getCanastaLocalSettingsFilePath() {
+	return getenv( 'MW_VOLUME' ) . '/config/LocalSettings.php';
+}
+
+if ( defined( 'MW_CONFIG_CALLBACK' ) ) {
+	// Called from WebInstaller or similar entry point
+
+	if ( !file_exists( getCanastaLocalSettingsFilePath() ) ) {
+		// Remove all variables, WebInstaller should decide that "$IP/LocalSettings.php" does not exist.
```
vedmaka, line 16: Nice trick!  👏

_sources/canasta/CanastaDefaultSettings.php
```diff
@@ -0,0 +1,73 @@
...
+			unset( $$k );
+		}
+		unset( $vars, $v, $k );
+		return;
+	}
+}
+// WebStart entry point
+
+// Check that user's LocalSettings.php exists
+$canastaLocalSettingsFilePath = getCanastaLocalSettingsFilePath();
```
vedmaka, line 28: Not a big deal, but it feels like this can be moved up so

https://github.com/WikiWorks/Canasta/pull/47/files#diff-f57eb0e3f924507019469c028a7783b875a27f117dfd5d51869cb272c224ca3dR15 

will use the variable value instead of calling the function?

jeffw16:
I tested it and this is awesome! LGTM! I give it my approval

jeffw16 (approved):

vedmaka (approved):

_sources/canasta/CanastaDefaultSettings.php
pastakhov: Thanks, you are right. I tried to not use variables, but then I noticed that it calls DefaultSettings.php and I have to remove all variables anyway.

jeffw16:
Testing seems to make it good, let's merge

PR #50, opened 2022-03-16
Updates msmtp configuration & configuration file location

vedmaka:
It was noticed that on debian the  `msmtp`  is looking for the configuration path at `/etc/msmtprc` and not `/etc/.msmtprc`


jeffw16 (approved):
LGTM

jeffw16:
That is strange that CI is not working on this PR...

vedmaka:
Dropped in favor of #62

PR #60, opened 2022-03-30
Changes default extensions and skins paths

vedmaka:
Modifies `$wgExtensionDirectory` , `$wgExtensionAssetsPath` , `$wgStyleDirectory` and `$wgStylePath` to use `user-` prefixed directories by default to make it look for skins and extensions in user mounted `user-skins` and `user-extensions` catalogue when `wfLoadExtension`  and `wfLoadSkin` is called


jeffw16 (approved):
LGTM

pastakhov (approved):

PR #61, opened 2022-03-31
Composer customisations and composer update at runtime

vedmaka:
* Allows to use `config/composer.local.json` file to install dependencies for user extensions
* Runs `composer update` on each container restart (preserving Canasta extensions and skins installed at build time) if `config/composer.local.json` is present
* Updates README
* Changes mount points for extensions and skins to `user-extensions` and `user-skins`

The last is also checked at the init script, the contianer won't run old mount points are discovered so this is a **breaking change**.


vedmaka:
Actually I think #60 need to be merged first and then I'll rebase this

tosfos:
I think we can merge that.

yaronkoren:
The new line 552 in the Dockerfile looks unusual:

COPY _sources/configs/composer.canasta.json $MW_HOME/composer.local.json

I'm not sure what this does, but won't it cause confusion, to rename composer.canasta.json to composer.local.json, when there is already a separate composer.local.json?

vedmaka:
@yaronkoren nope, the Canasta default composer file being copied is intended to act as a root `composer.local.json` within the container and user mounted file is to `/mediawiki/config/composer.local.json` and then included by the root one

if you think having a `composer.canasta.json` inside a contianer would look nicier - the above can be modified to `composer.local.json` <- `composer.canasta.json` <-  `/mediawiki/config/composer.local.json`  scheme

PR #65, opened 2022-04-04
Order the files included in the settings/ directory

### LINKED ISSUE #64: Order the files included in the settings/ directory
Currently, we are globbing the `settings/` directory to load in all of the files under that directory. Globbing the `settings/` directory does not guarantee any ordering of the files. That can be dangerous, as the files would be loaded in a nondeterministic order, which can change the functionality of the wiki without warning. It would be best to load them by alphabetical order of the file names.

Here is how globbing currently works: https://github.com/CanastaWiki/Canasta/blob/7cc8b1a045f5ee5ce0c687d62cd44535dad3b2fe/_sources/canasta/CanastaDefaultSettings.php#L73

abhi-bhatra:
Fixes #64


abhi-bhatra:
Hi @jeffw16 Do check this out, and guide me if it solves the purpose !

jeffw16 (approved):
Thanks! LGTM, but I would like one more approval from either @pastakhov or @vedmaka before merging.

tosfos:
WikiTeq will review this soon

pastakhov:
It is already alphabetically sorted see https://github.com/CanastaWiki/Canasta/issues/64#issuecomment-1140485435

PR #69, opened 2022-04-08
Change CanastaUtils include location

jeffw16:
`CanastaUtils.php` should be included from `CanastaDefaultSettings.php` because some people may want to remove this file from their Canasta derivative image. Our requirement for people is to never edit `sources/canasta/LocalSettings.php` and only edit `sources/canasta/CanastaDefaultSettings.php` for official Canasta support.

Does this change break anything?


pastakhov:
I didn't test it, but I think this can't break anything

PR #71, opened 2022-04-08
Fixes Bootstrap extension patch

vedmaka:
The git patch was not applying correctly, replaced with `patch`


pastakhov (approved):
I approve the change but I don't like the patch of bootstrap and idea in general

Issue #55, opened 2022-03-28
Tag images with Canasta releases too


jeffw16:
For instance, Canasta 1.0.0

vedmaka:
Apparently this was already done in https://github.com/CanastaWiki/Canasta/pull/79 thus closing

PR #79, opened 2022-04-14
Adding version file

jeffw16:
This file will represent the version of the Canasta image (and should influence the tag on GHCR too).


jeffw16:
I have no idea why the build would be failing but it has nothing to do with this PR, which literally adds a file called `VERSION` with the value `1.0.0`.

yaronkoren (approved):

vedmaka:
Closes #55

PR #87, opened 2022-04-28
Update Cargo, Page Forms versions

yaronkoren:


tosfos:
Thanks for the update! I know the Stability specs mention "Extensions will be tied to specific git commits and will rarely be updated." Is there a specific reason to update these extensions? Is there any reason to think they could be introducing any bugs?

jeffw16:
> Is there a specific reason to update these extensions?

Bug fixes/security vulnerability fixes for sure.

I think in general, it is OK for newer versions of Canasta to progressively feature newer and newer versions of extensions. If people want an older version of Page Forms, they can use an older version of Canasta.

Perhaps it's time to make a development trunk branch and start spinning off tags for newer releases of Canasta. This PR would be a perfect candidate to be merged into the development trunk branch rather than `master`/`main`.

yaronkoren:
As I said, it's both. It's hard to say for sure whether any new version is more stable than a previous version - I'd like to think the answer is yes, but I can't say for sure.

Let me also note that this discussion feels rather trivial at the moment, since Canasta isn't fully working yet! For now, probably any version is fine. Once Canasta is fully working, it'll probably make more sense to talk about an ideal upgrade strategy for these often-updated extensions.

tosfos:
Yaron, based on our discussions should we abandon this?

yaronkoren:
Sure, yeah. I was thinking we could just keep it until the MW 1.39 version of Canasta comes out, but by then there will probably be different versions of these two extensions to use.

PR #95, opened 2022-05-26
Remove Favorites extension

### LINKED ISSUE #94: Fix or remove the Favorites extension
Favorites is a MediaWiki extension contained in Canasta that lets you mark different wiki pages as your "favorites", and then view the list of your favorites:

https://www.mediawiki.org/wiki/Extension:Favorites

As you can see, last month two big warnings were added to the page: that the extension is incompatible with MW 1.33+, and that it's unmaintained.

The incompatibility is because it calls a method, APIBase::dieUsageMsg(), that was removed in MW 1.32.  And due to some very strange coding, this method seems to always get called - not just on errors. (?) See here:

https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/Favorites/+/refs/heads/master/includes/api/ApiFavorite.php#33

On top of that, Favorites also uses a hook, InternalParseBeforeSanitize, that was deprecated in MW 1.35:

https://www.mediawiki.org/wiki/Manual:Hooks/InternalParseBeforeSanitize

So, I see two options for this extension: improve it ourselves, or remove it from Canasta. (There's also the option of just waiting and hoping that someone else will fix it, but that seems optimistic.) Personally, I vote for removing it - I don't really see the point of marking certain pages as your favorites, and for users who really want to do that, they could always use browser bookmarks or something along those lines.

yaronkoren:
Closes #94.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -263,10 +263,6 @@ RUN set -x; \
 	&& git clone --single-branch -b master https://gerrit.wikimedia.org/r/mediawiki/extensions/ExternalData $MW_HOME/extensions/ExternalData \
 	&& cd $MW_HOME/extensions/ExternalData \
 	&& git checkout -q 64785b7e2134121d84a77edde9daab5db040e97a \
-	# Favorites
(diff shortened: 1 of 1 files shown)


jeffw16:
Closes #94

jeffw16 (approved):
LGTM

tosfos (approved):

PR #110, opened 2022-06-19
Adds `poppler-utils` as an `xpdf` replacement for `PdfHandler`

vedmaka:
PdfHandler was not functioning due to `pdfinfo` and `pdftotext` tools missing. This patch installs `poppler-utils` as a replacement for `xpdf-utils` for Debian.


jeffw16 (approved):
Glad this change is being made, LGTM

tosfos (approved):

Issue #114, opened 2022-06-24
API for accessing MediaWiki info set in LocalSettings.php, etc?


jeffw16:
Canasta currently has no way to access the info stored in `LocalSettings.php`, so it can't, for instance, detect the database name. Can we make an API for doing this?

yaronkoren:
I agree that this would be useful, but, rather than API action(s), I think this would be best implemented as a command-line script inserted within MediaWiki - it seems like it would be more secure, so that regular users couldn't access this information. But maybe by "API" you didn't necessarily mean the literal MW API.

jeffw16:
> But maybe by "API" you didn't necessarily mean the literal MW API.

Correct, APIs come in many forms, and I should've been more clear that this should not be a web API.

> I think this would be best implemented as a command-line script inserted within MediaWiki 

I completely agree!

> it seems like it would be more secure, so that regular users couldn't access this information

Completely agreed; in fact, I see no reason why a regular user should be using this proposed API.

yaronkoren:
This is done now.

PR #116, opened 2022-06-27
Create displayWikiInfo.php

yaronkoren:
As recommended in issue #114.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -677,6 +677,7 @@ COPY _sources/configs/php_error_reporting.ini _sources/configs/php_upload_max_fi
 COPY _sources/configs/php_max_input_vars.ini _sources/configs/php_max_input_vars.ini /etc/php/7.4/apache2/conf.d/
 COPY _sources/configs/php_timeouts.ini /etc/php/7.4/apache2/conf.d/
 COPY _sources/scripts/*.sh /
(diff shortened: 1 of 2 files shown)


yaronkoren:
I've tested this in a regular wiki instance but not within Canasta; so I'm guessing that at least the path on line 32 is incorrect.

Dockerfile
```diff
@@ -677,6 +677,7 @@ COPY _sources/configs/php_error_reporting.ini _sources/configs/php_upload_max_fi
 COPY _sources/configs/php_max_input_vars.ini _sources/configs/php_max_input_vars.ini /etc/php/7.4/apache2/conf.d/
 COPY _sources/configs/php_timeouts.ini /etc/php/7.4/apache2/conf.d/
 COPY _sources/scripts/*.sh /
+COPY _sources/scripts/*.php $MW_HOME/maintenance/
```
jeffw16, line 680: Should we place this in an extension in something like `$MW_HOME/extensions/Canasta/maintenance/`?

like this https://github.com/samwilson/RenderThumbnails/

jeffw16 (approved):
Approving but see comments

Dockerfile
jeffw16: Let's worry about this later

Issue #120, opened 2022-07-07
Add CookieWarning extension


yaronkoren:
I think we should add the CookieWarning extension to Canasta:

https://www.mediawiki.org/wiki/Extension:CookieWarning

It adds one of those now-ubiquitous "This site uses cookies. Is that OK?" dialogs to the wiki. It's a widely-used extension, seems to work well, and (to the extent that I understand the law) offers functionality needed by anyone running a public Canasta-based site in the EU.

tosfos:
I'd rather change the law. But if that's not an option, we should add this extension.

jeffw16:
I also think we should add it.

yaronkoren:
I agree that it would be better to get rid of GDPR, but until we can do that, I'm glad you both agree that this is the next best thing! I just created a patch for it.

PR #121, opened 2022-07-11
Add CookieWarning extension

yaronkoren:
Resolves isssue #120.


jeffw16:
Can we change the destination branch of this PR to release-1.1.2? (It doesn't exist yet but we should make it.)

jeffw16:
I think the build is failing because the latest changes from `release-1.1.1` should be merged in first to fix an old version of Apache acting up.

PR #124, opened 2022-07-13
Upgrades debian packages

vedmaka:
* Upgrades `apache2` from `2.4.53-1~deb11u1` to `2.4.54-1~deb11u1`
* Upgrades `gpg` from `2.2.27-2+deb11u1` to `2.2.27-2+deb11u2`
* Updates `gnupg` from `2.2.27-2+deb11u1` to `2.2.27-2+deb11u2`
* Updates `ffmpeg` from `7:4.3.3-0+deb11u1` to `7:4.3.4-0+deb11u1`

once merged please rebase and merge #125

PR #125, opened 2022-07-13
Fixes root `debian:bullseye` image version to `11.4` (latest)

vedmaka:
Depends-On: #124


jeffw16 (approved):
LGTM, I'm glad we are making this change

PR #127, opened 2022-07-22
Makes use of `gateway.docker.internal` to detect host address

### LINKED ISSUE #126: Hardcoded docker host address
There is a need for container to connect to Docker host for:

1. Sending emails
2. Do VisualEditor Parsoid API calls

Right now there is a hard-coded default `172.17.0.1` [[1]](https://github.com/CanastaWiki/Canasta/blob/master/_sources/configs/msmtprc#L9) [[2]](https://github.com/CanastaWiki/Canasta/blob/master/_sources/scripts/run-apache.sh#L65) address for Docker host being used, this works fine most of the time, however, considering default bridge network may vary either due to changes made to Docker `daemon.json` configuration file on target host or due to hitting limit of subnets occupied.

It'd be better to have this address detected automatically, fortunately, starting Docker `20.04` this is possible via `--add-host=host.docker.internal:host-gateway` param or `extra_hosts` attribute in a Compose file

vedmaka:
* Updates `run-apache.sh` to use `gateway.docker.internal` as host address and fall back to `172.17.0.1` if not available
* Updates `msmtprc` to contain replacement pattern to be processed via `run-apache.sh`
* Updates `$wgCdnServersNoPurge` to include all possible docker networks

Fixes #126


_sources/canasta/CanastaDefaultSettings.php
```diff
@@ -66,6 +66,8 @@
 $wgUseCdn = true;
 $wgCdnServersNoPurge = [];
 $wgCdnServersNoPurge[] = '172.16.0.0/12';
+$wgCdnServersNoPurge[] = '192.168.0.0/16';
+$wgCdnServersNoPurge[] = '10.0.0.0/8';
```
jeffw16, line 70: What's the purpose of adding these subnets?

jeffw16 (approved):
LGTM

_sources/canasta/CanastaDefaultSettings.php
pastakhov: We know that one of them will be used, so we added them all because it is better to add unused networks than it will be broken when it uses not added network...

pastakhov (approved):

PR #131, opened 2022-08-13
Remove packages versions fixings

vedmaka:
Apparently the idea to use strict packages versions was not so good idea, it seems that packages being updated in Debian repo with old versions removes, this leads to unexpected fails on builds not related to PR changes, eg: https://github.com/CanastaWiki/Canasta/runs/7716246944?check_suite_focus=true

It should be generally safe to remove versions fixings and just always stay up to date on the packages


jeffw16 (approved):
Agreed, LGTM

pastakhov (approved):

PR #135, opened 2022-10-03
Canasta 1.2.0

### LINKED ISSUE #133: Get rid of cfLoad...() functions
Some people (maybe many) will be happy about this, but: I have been convinced that switching away from the cfLoadSkin() and cfLoadExtension() functions is the best course of action for Canasta. I admit that it's not actually the technical arguments that convinced me, but just the seeming consensus on the other side - both among developers and regular users. So, the new plan of action is to keep the four directories (now called extensions/, skins/, canasta-extensions/ and canasta-skins/ - maybe this will change as well?), but to just use wfLoadExtension() and wfLoadSkin() to do that. It seems like there are two main possible approaches for this:

1) Just symlink every built-in extension and skin, so that the main extensions/ and skins/ directories start out with a ton of symlinks each.
2) Something like @jeffw16's suggested approach, of having Canasta, on startup, figure out which extension and skin is contained in which directory, and thus create something like internal symlinks, to enable wfLoadSkin() and wfLoadSkin() to always call the "right" version - the user-installed one if it exists, and the Canasta-packaged one otherwise. (Jeff - please correct me if I'm wrong, and feel free to offer more information in any case.)

No matter what approach is taken, there will still be issues if an admin installs their own version of a Canasta-packaged skin or extension, and then wants to toggle back and forth between the two versions, maybe for testing purposes, but clearly this problem does not outweigh the benefits of getting rid of the cfLoad... functions, for most people.

jeffw16:
Removes cfLoad and utilizes symlinking method to detect which extension/skin asset to load. Fixes #133

Upgrades to 1.35.8.


yaronkoren:
This looks good to me. It's great that the new symlinking approach can be done with a relatively small amount of code, and it's also great that the old cfLoad.. extensions are being kept, to deliver warnings. I probably would have put the MW version change into a separate patch, but I can see why it's integrated into this one, to make it a real "omnibus" change.

yaronkoren (approved):

PR #136, opened 2022-10-07
Upgrades SimpleBatchUpload from `1.0` to `1.8.x`

vedmaka:
The patch bumps SimpleBatchUpload version to `1.8.x`, it was tested on WikiTeq-Canasta and worked well there


yaronkoren (approved):

jeffw16:
I'm not sure why I missed this in my emails, but next time could we please not merge directly into master and instead merge into a versioned release branch?

PR #161, opened 2022-11-27
Fully install EmailAuthorization extension

### LINKED ISSUE #155: Install EmailAuthorization extension completely
It turns out that EmailAuthorization doesn't get fully installed by Canasta, because it requires calling "git submodule update --init":

https://www.mediawiki.org/wiki/Extension:Email_Authorization

I would think the best way to fix this is to add this call to the Dockerfile, as is already done for VisualEditor:

https://github.com/CanastaWiki/Canasta/blob/master/Dockerfile#L92

yaronkoren:
Fixes #155 .


Dockerfile
```diff
@@ -89,6 +89,9 @@ RUN set -x; \
 	&& git submodule update --init --recursive \
     # VisualEditor
     && cd extensions/VisualEditor \
+    && git submodule update --init \
+    # EmailAuthorization
+    && cd ../EmailAuthorization \
```
jeffw16, line 94: I think this should be `cd ../../EmailAuthorization`

jeffw16 (changes requested):

Dockerfile
yaronkoren: That might explain the error message, but are you sure about this? Does calling "git submodule update" bring one to a subdirectory?

jeffw16: Calling `git submodule update` should not change the working directory.

yaronkoren: Okay. In that case, this current code is fine, no?

yaronkoren: The code has changed, so this has become irrelevant anyway.

yaronkoren:
Okay, after five attempts, I think I finally got these two extra lines to work. :) @jeffw16 - thoughts?

Dockerfile
```diff
@@ -527,6 +524,15 @@ RUN set -x; \
...
 
+# Get all Git submodules
+RUN set -x; \
+	# VisualEditor
+	cd $MW_HOME/extensions/VisualEditor \
+	&& git submodule update --init \
+	# EmailAuthorization
+	&& cd $MW_HOME/extensions/EmailAuthorization \
+	&& git submodule update --init
+
```
vedmaka, line 535: While it's definitely good to have logically separated parts grouped into own `RUN` commands we have to take into account Docker schrodinger limit [1] [2] on maximum number of layers Dockerfile can contain (each `RUN` creates a new layer). So we better to be careful with that :)

1. https://github.com/docker/docs/issues/8230#issuecomment-475278273
2. https://stackoverflow.com/questions/39382518/whats-the-reason-for-the-42-layer-limit-in-docker/54936566#54936566

yaronkoren: Oh, I didn't know about that. Well, I think the Dockerfile currently contains 23 "RUN" calls - and this patch would make it 24. But some of those RUN calls are for patches, and I think the MW 1.39 of Canasta (coming soon!) will have 6 fewer patches than there currently are. So once that is done, we will be down to 18. That seems fine, no?

jeffw16 (approved):
LGTM

PR #163, opened 2022-12-02
Update ApprovedRevs, Cargo, PageForms, VEForAll to the 1.39 version

yaronkoren:
…upports MW 1.39


jeffw16 (approved):
I haven't tested any of these extensions, but approved

PR #164, opened 2022-12-02
Remove now-unnecessary "path patches"

yaronkoren:
I would think that, with the removal of the whole cfLoad...() system, these changes to the default handling are now unnecessary.


jeffw16 (approved):
LGTM

PR #168, opened 2022-12-04
Remove patches that are unnecessary for MW 1.39

yaronkoren:
These four patches are unnecessary for the latest code: the patches to CommentStreams, Semantic MediaWiki and Semantic Result Formats were all added already to their respective extensions, while the patch for core MediaWiki, done in order to get the GoogleAnalyticsMetrics extension working, is no longer necessary due to a recent fix to GoogleAnalyticsMetrics.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -538,12 +538,6 @@ RUN set -x; \
     tar -xvf /tmp/GTag*.tar.gz -C $MW_HOME/extensions \
     && rm /tmp/GTag*.tar.gz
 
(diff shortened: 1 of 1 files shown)


yaronkoren:
(This patch deletes one extra file that it shouldn't - [core-fix-for-curl-a2f60bb.diff](https://github.com/CanastaWiki/Canasta/pull/168/commits/aef7c1020ca903ad4155f5c2c424e91ae43454f0) - but I think that's alright, because that file will hopefully have been deleted anyway by the time this PR is merged.)

PR #173, opened 2022-12-07
Add patch for removing errorbox from SMW extension

### LINKED ISSUE #143: Remove "Semantic MediaWiki was installed but not enabled on this wiki" message from Special:Version
Semantic MediaWiki displays the following "error message" at the top of Special:Version if it's installed but not enabled:

> Semantic MediaWiki was installed but not enabled on this wiki. Please consult the [help page](https://www.semantic-mediawiki.org/wiki/Extension_registration) for instructions and further assistances [sic].

Regardless of whether or not this error message is ever a good idea, it's bad in Canasta - where, due (I think) to the way the Validator extension gets installed via Composer, SMW _always_ thinks it's installed - even for users who have no interest in using it. I see two main options here: patch SMW to get rid of the error message, or add some CSS to hide "errorbox" divs within the Special:Version page. (I think SMW is the only extension that adds an error message to Special:Version.)

A third option, of course, is to attempt to get the SMW developers to themselves remove this error message "upstream", but I don't know how likely that is to work, and it would probably be too later for this MW 1.39 release anyway. (Though it's a good idea, no matter what.)

sanjay-thiyagarajan:
Fixes #143


jeffw16:
Sanjay makes some awesome contributions to Canasta, this one included, so it's really annoying how Sanjay's changes won't build because he doesn't have contributor permissions to Canasta. Anyone opposed to granting him contributor access? I'd like to add it to him ASAP so we can at least verify this builds properly, then I'll be able to do a CR.

yaronkoren:
No objections here! It makes a lot of sense to include him.

yaronkoren:
@jeffw16 - by the way, I think it makes sense to make this a patch for the MW 1.39 branch, not the 1.2.2 branch, because the MW 1.35 version of the relevant SMW file ([BeforePageDisplay.php](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/MediaWiki/Hooks/BeforePageDisplay.php)) is slightly different from the MW 1.39 version, so I think this patch won't work with the earlier SMW version.

yaronkoren:
@sanjay-thiyagarajan - it turns out that this problem might be happening because the pull is from another repository. (Well, that plus some weird permissions issues.) Could you close/abandon this PR, and create a new one instead, directly in this repo?

PR #175, opened 2022-12-08
Removes unnecessary patch for Refreshed skin

### LINKED ISSUE #166: Update or remove Refreshed skin patch
Canasta contains a patch to the Refreshed skin, by @vedmaka, to comment out HTML-loading errors so that they only show up in the HTML source of the page, and not the screen:

https://github.com/CanastaWiki/Canasta/blob/master/_sources/patches/skin-refreshed.patch

The Dockerfile contains a @todo to merge this patch directly into the Refreshed skin, but evidently that didn't happen over the last year, because that part of the Refreshed skin looks the same:

https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/skins/Refreshed/+/refs/heads/master/includes/RefreshedTemplate.php#399

The line number has changed, though - so this patch needs to be either updated, or removed, for the 1.39 version of Canasta. Personally, I don't see the need to comment out major errors like this, but maybe there's something I don't know about.

vedmaka:
Fixes #166

This patch is a WikiTeq image legacy and should not be in Canasta image, thus removed


pastakhov (approved):

jeffw16:
Could we please avoid merging into the master/main branch?

yaronkoren:
Oh, sorry - I didn't think about that. Some changes (like bug fixes) are fine to go into master, but this was not one of them.

jeffw16:
Bug fixes are unfortunately not okay to go into the main branch either (unless they are security bugs) because it provides inconsistent versioning. Who knows, a bug fix might break things further.

jeffw16:
Yes, main branch should only have changed made to it if the "VERSION" file is also updated

PR #184, opened 2022-12-12
Remove "VectorGeneratedSkinData" hook patch

### LINKED ISSUE #167: Update or remove Vector skin's "VectorGeneratedSkinData" patch
Canasta contains a patch to the Vector skin, by @pastakhov, that adds the hook "VectorGeneratedSkinData" to the Vector skin's VectorTemplate.php file:

https://github.com/CanastaWiki/Canasta/blob/master/_sources/patches/skin-vector-addVectorGeneratedSkinDataHook.patch

This patch will not work any more with the MW 1.39 version of Canasta, because the file VectorTemplate.php no longer even exists, as far as I can tell. So the patch needs to be either updated or removed - depending, I assume, on whether that hook is still being used.

yaronkoren:
Fixed #167.


jeffw16 (approved):
LGTM

PR #187, opened 2022-12-15
Add header warning to some special pages if $wgSMTP not set

### LINKED ISSUE #145: Clarify to users that $wgSMTP is required for mailing
**Updated task**: Make it clear to admins that they will need to use an email service (i.e., by setting $wgSMTP) in order to have the wiki send emails. Information should be added in the following places:

- On the website, maybe in a new "Known issues" section
- In the CLI installer, as text that shows up at some point during installation
- On the wiki itself - in a warning message box that shows up, maybe at the top of the page, if $wgSMTP is not set, on some set of pages and for some set of users.

yaronkoren:
Fixes #145.


yaronkoren:
This got too complex, so I'm abandoning it and replacing it with #194.

PR #194, opened 2022-12-22
Add header warning to some special pages if $wgSMTP not set

### LINKED ISSUE #145: Clarify to users that $wgSMTP is required for mailing

yaronkoren:
Fixes #145.

A new version of the now-abandoned #187.

### DIFF
--- a/_sources/canasta/CanastaUtils.php
+++ b/_sources/canasta/CanastaUtils.php
@@ -51,4 +51,4 @@ $wgHooks['SiteNoticeAfter'][] = function ( &$siteNotice, Skin $skin ) {
 
 	$siteNotice .= '<div class="warningbox" style="font-size: larger;">Please note that mailing does not currently work on this wiki, because Canasta requires <a href="https://www.mediawiki.org/wiki/Manual:$wgSMTP">$wgSMTP</a> to be set in order to send emails.</div>';
 	return true;
-}
+};
--- a/_sources/canasta/CanastaUtils.php
+++ b/_sources/canasta/CanastaUtils.php
@@ -22,3 +22,30 @@ function cfLoadSkin( $skinName ) {
 	echo "The following skin was loaded with cfLoadSkin, but needs to be loaded with wfLoadSkin instead: $skinName";
 	die();
 }
+
+/**
+ * Not exactly a utility function, but - show a warning to users if $wgSMTP is not set.
+ */
+$wgHooks['SiteNoticeAfter'][] = 'showSMTPWarning';
+function showSMTPWarning( &$siteNotice, Skin $skin ) {
+	global $wgSMTP;
+
+	if ( $wgSMTP !== false ) {
+		return true;
+	}
+	$title = $skin->getTitle();
+	if ( !$title->isSpecialPage() ) {
+		return true;
+	}
+	$specialPage = MediaWiki\MediaWikiServices::getInstance()
+		->getSpecialPageFactory()
+		->getPage( $title->getText() );
+	$canonicalName = $specialPage->getName();
+	// Only display this warning for pages that could result in an email getting sent.
+	if ( !in_array( $canonicalName, [ 'Preferences', 'CreateAccount', 'Emailuser' ] ) ) {
+		return true;
+	}
+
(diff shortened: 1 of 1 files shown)


tosfos:
Shouldn't it check if $wgEnableEmail is true?

yaronkoren:
How interesting - I didn't know about that setting. I looked it up, and also saw this, which is specifically for Special:Emailuser:

https://www.mediawiki.org/wiki/Manual:$wgEnableUserEmail

So probably the code should be checking both.

_sources/canasta/CanastaUtils.php
```diff
@@ -22,3 +22,34 @@ function cfLoadSkin( $skinName ) {
 	echo "The following skin was loaded with cfLoadSkin, but needs to be loaded with wfLoadSkin instead: $skinName";
 	die();
 }
+
+/**
+ * Not exactly a utility function, but - show a warning to users if $wgSMTP is not set.
+ */
+$wgHooks['SiteNoticeAfter'][] = 'showSMTPWarning';
+function showSMTPWarning( &$siteNotice, Skin $skin ) {{
```
pastakhov, line 30: I think there should be only one open brace {

and do we really need a name for the function there?  why can't we use anonymous function like:
```
$wgHooks['SiteNoticeAfter'][] = function ( &$siteNotice, Skin $skin ) {
...
};
```

_sources/canasta/CanastaUtils.php
```diff
@@ -22,3 +22,34 @@ function cfLoadSkin( $skinName ) {
...
+	// Only display this warning for pages that could result in an email getting sent.
+	$specialPagesWithEmail = [ 'Preferences', 'CreateAccount' ];
+	if ( $wgEnableUserEmail ) {
+		$specialPagesWithEmail[] = 'Emailuser';
+	}
+	if ( !in_array( $canonicalName, $specialPagesWithEmail ) ) {
+		return true;
+	}
+
+	$siteNotice .= '<div class="warningbox"><big>Please note that mailing does not currently work on this wiki, because Canasta requires <a href="https://www.mediawiki.org/wiki/Manual:$wgSMTP">$wgSMTP</a> to be set in order to send emails.</big></div>';
```
pastakhov, line 53: the  big tag is deprecated https://developer.mozilla.org/en-US/docs/web/html/element/big

_sources/canasta/CanastaUtils.php
```diff
@@ -22,3 +22,34 @@ function cfLoadSkin( $skinName ) {
...
 }
+
+/**
+ * Not exactly a utility function, but - show a warning to users if $wgSMTP is not set.
+ */
+$wgHooks['SiteNoticeAfter'][] = 'showSMTPWarning';
+function showSMTPWarning( &$siteNotice, Skin $skin ) {{
+	global $wgSMTP, $wgEnableEmail, $wgEnableUserEmail;
+
+	if ( $wgEnableEmail == false || $wgSMTP !== false ) {
```
pastakhov, line 33: by default $wgSMTP is null (or at least can be null) and in this case it won't works as expected.   I would simplify this as:
```
if ( !$wgEnableEmail || $wgSMTP  ) {
    return true;
}
```

pastakhov (changes requested):

yaronkoren:
@pastakhov - thanks for these fix suggestions! I think I added all of them. The double curly bracket was just a copying error. $wgSMTP defaults to false, not null, but you're right that we might as well check for everything.

jeffw16:
I will refrain from providing a review until @pastakhov has had a chance to confirm his comments have been fixed to his satisfaction.

pastakhov (approved):
I'm satisfied, at least it does not break wiki now.
just a few notices:
* since MediaWiki 1.22 (in 2015) no need to return true in hooks
* I don't see styles for "warningbox" class in 1.39, I would use Html::warningBox( 'my message' ) instead;

yaronkoren:
@pastakhov - thanks for the comments! I'm embarrassed that I never actually checked this code on an MW 1.39 wiki. The "warningbox" class indeed did not work. I've fixed the code, and tested it, so now it should work better.

PR #195, opened 2022-12-24
Add patch for removing errorbox from SMW extension

sanjay-thiyagarajan:


jeffw16 (approved):
LGTM

pastakhov (changes requested):
It seems we don't need this patch for 1.39, it uses "mediawiki/semantic-media-wiki": "^4.0.0" and it is moving away from enableSemantics and does not show the error box as I see this in https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/86bcff16f63975073235456c8bd2d6fc9d270507/includes/SemanticMediaWiki.php#L40

yaronkoren:
@pastakhov - oh, how interesting, it does look like SMW's informAboutExtensionAvailability() function no longer does anything, judging from the code.

Does anyone have access to a wiki running a version of SMW whose code is from less than a year ago (this change was made [about a year ago](https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/f1502d1d4b053ecdd0a12963145baf4bc586c504)), and would be willing to briefly disable SMW in their LocalSettings.php, to see if the warning in Special:Version no longer appears? It would be good to know for sure.

pastakhov:
I don't see the warning in my 1.39 test wiki, but it is better to check it with somebody else for sure.

yaronkoren:
I just confirmed that this code in fact does not run with MW 1.39. It should just be removed from SMW, but that's not our department. 😄 @sanjay-thiyagarajan - sorry about the unnecessary work you did for this. I'm closing this patch.

PR #200, opened 2022-12-27
Delete backup_scripts/, backups/ and kubernetes/ directories (!)

### LINKED ISSUE #197: Remove backup_scripts/, backups/ and kubernetes/ directories
This was originally just a comment to issue #113, but it seems to be bigger than that, so I think a separate issue makes more sense. The backup-scripts/ directory currently contains two shell scripts that I think don't work (just by looking at them), and I assume are never used. One of the scripts, backup.sh, is called by a cron job defined in the file [database_backup_cron.yaml](https://github.com/CanastaWiki/Canasta/blob/master/kubernetes/database_backup_cron.yaml), but I assume that cron job is never called either. For that matter, I would guess that none of the YAML files in the directory https://github.com/CanastaWiki/Canasta/tree/master/kubernetes are used - after all, we have a separate repository for all the Kubernetes handling, at https://github.com/CanastaWiki/Canasta-Kubernetes.

The backup scripts save their files to the directory backups/, so I assume that directory, too, is never used. So, it seems to me that all three of these directories should be deleted.

Let me clarify that I'm not against having backup capability - it would be great to have simple backup capability like this, especially for people who don't want to use Restic. (And of course, I'm also not against Kubernetes support.) But if we just have scripts and directories that don't work and are never called, and they are not going to get fixed any time soon, then we might as well just delete them now, to avoid confusing users. I'm not worried about losing important code here, by the way - it's not that hard to backup a wiki, and mediawiki.org has [a variety of backup scripts](https://www.mediawiki.org/wiki/Manual:Backing_up_a_wiki#Scripts) that we could potentially use, if/when backup capability is added.

Any thoughts? CC @jeffw16, @pastakhov, @vedmaka

yaronkoren:
These are all unused - fixes #197


jeffw16 (approved):
LGTM

PR #201, opened 2022-12-27
Remove SocialProfile patches

### LINKED ISSUE #198: Remove SocialProfile patches
Canasta currently includes two patches for the SocialProfile extension: social-profile-REL1_35.44b4f89.diff and SocialProfile-disable-fields.patch. The first is a security fix that was added to SocialProfile a while ago, while the second is a Canasta-specific addition (it adds the page Special:ManageFields, and makes some related changes). Regardless, they should both be removed, since they apply only to the MW 1.35 version of SocialProfile. If there's going to be a new version of the second patch, I would think it should be created separately.

yaronkoren:
Neither is called any more with MW 1.39 - and they wouldn't work even if they were called.

Fixes #198.


jeffw16:
Is it really true it doesn't work in 1.39 anymore? Where does it say that this is the case?

yaronkoren:
@jeffw16 - well, the "REL1_35" patch was already added directly to SocialProfile a while ago. As for the "disable-fields" patch - I haven't tried it, but I assume, given the large size of the patch, that there's no way it would work with the MW 1.39 version of SocialProfile - even a single line change to any of the files in SocialProfile between then and now would probably break it.

Issue #159, opened 2022-11-27
Automatically update extension and skin symlinks when changes occur


yaronkoren:
As noted [here](https://github.com/CanastaWiki/Canasta/pull/158#issuecomment-1327721512) by @pastakhov, it would be great to have some mechanism to update the extension and skin symlinks when things are added to, or removed from, the user-extensions/ and user-skins/ directories. Using the Linux inotifywait command, as pastakhov suggested, seems like the best way to do this.

PR #206, opened 2023-01-16
Add the ability to detect changes on the folders extensions and skins and add appropriate symlinks.

gitmapd:
- Inotifywait detects activity changes in the folders user-extensions and user-skins.
- Adds symlinks from the those folders to the appropriate correspondent folders, user-extensions to extensions and user-skins to skins. As a fallback, if the administrator wants to go back to the default extension provided by canasta-extensions folder it will then revert back.
- Dockerfile had to be updated because inotifywait isn't installed by default, so inotify-tools package add to be added to be installed.
- All scripts are called from Run-apache.sh, so  a call to inotifywait.sh was  added to that script.


yaronkoren:
Fixes #159 .

_sources/scripts/inotifywait.sh
```diff
@@ -0,0 +1,16 @@
...
+extensions="$MW_HOME/extensions"
+canexts="$MW_HOME/canasta-extensions"
+
+userskins="$MW_HOME/user-skins"
+skins="$MW_HOME/skins"
+canskins="$MW_HOME/canasta-skins"
+
+#Detects changes on the user-extensions and user-skins directories and symlinks to each directory skins and extensions else if removed it defaults back to
+#canasta-extensions or canasta-skins
+inotifywait -m -e create,moved_to,delete,moved_from --format '%e:%f%0' -- "$userexts"| while IFS=: read -r event file; do case $event in CREATE,ISDIR|MOVED_TO,ISDIR) ln -rsft "$extensions" -- "$userexts"/"$file" ;; DELETE,ISDIR|MOVED_FROM,ISDIR) echo "event: ${event}   file: ${file}"; ln -rsft "$extensions" -- "$canexts"/"$file" || rm -- "$file"; esac done
```
jeffw16, line 14: Oh my goodness. I'm sorry, I simply cannot read this code. Would it be possible to please reformat it into something that isn't one line?

jeffw16 (changes requested):

PR #214, opened 2023-02-20
 Add monitoring for extensions and skins modifications.

gitmapd:
- Inotifywait detects activity changes in the folders user-extensions and user-skins.
- Adds symlinks from the those folders to the appropriate correspondent folders, user-extensions to extensions and user-skins to skins. As a fallback, if the administrator wants to go back to the default extension provided by canasta-extensions folder it will then revert back.
- Dockerfile had to be updated because inotifywait isn't installed by default, so inotify-tools package add to be added to be installed.
- All scripts are called from Run-apache.sh, so  a call to inotifywait.sh was  added to that script.


yaronkoren:
Replaced with #215, which passes validation.

PR #215, opened 2023-02-22
Add monitoring for extension and skins modifications (patch by @gitmapd)

yaronkoren:
His attempts to get this patch through kept failing due to 403 errors during validation, so we're seeing if I can get it through instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ RUN set x; \
 	&& aptitude -y upgrade \
 	&& aptitude install -y \
 	git \
+	inotify-tools \
 	apache2 \
 	software-properties-common \
 	gpg \
--- a/_sources/scripts/monitor-directories.sh
+++ b/_sources/scripts/monitor-directories.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+set -x
+
+userexts="$MW_HOME/user-extensions"
+extensions="$MW_HOME/extensions"
+canexts="$MW_HOME/canasta-extensions"
+
+userskins="$MW_HOME/user-skins"
+skins="$MW_HOME/skins"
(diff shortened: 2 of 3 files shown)


yaronkoren:
Fixes #159.

jeffw16 (approved):
I have no comments, so I'll approve this with the proviso that I haven't personally tested this.

vedmaka:
apparently this patch requires a follow up to be done in some future to support for extensions requiring either `update.php` script to be ran and/or having Composer dependencies

jeffw16:
Why is it overkill to run `update.php` and `composer update` both each time?

yaronkoren:
You're right that it might not be overkill at all - but on the other hand, thinking about it again now, I might have been correct, but for the wrong reason, that these calls should be avoided. The problem with update.php, if I'm thinking about it correctly, is not that it would be too costly to call, but that it actually wouldn't work at all: it will only make the relevant DB changes once the extension is enabled - but at the time this script runs, the extension has only been added, not enabled, so update.php will do nothing.

As for "composer update" - I don't know if it requires enabling the extension, but even if it does, the code could instead just go into the extension's directory and call "composer install", so that part is fine. However, it doesn't seem that useful: this would only be relevant for extensions that are installed via a two-step method (Git or download, then Composer), which I think is a relatively small number; and it would just save calling one command, which presumably the admin knows they need to call anyway, if they're following the installation instructions. Plus, maybe there's a chance that this extension requires calling something other than a plain "composer install" (or "composer install --no-dev"), so calling it automatically could potentially even mess up the installation. (I'm just speculating - maybe that's not possible.) Anyway, I think it would be only minimally useful.

jeffw16:
Oh right, makes sense. Great points Yaron!

Issue #35, opened 2022-02-22
Use GitHub mirrors instead of Gerrit


jamesmontalvo3:
Recommend using GitHub mirrors for WMF resources. See https://github.com/enterprisemediawiki/meza/pull/1061

jamesmontalvo3:
I suppose it's not worth the effort until it presents an issue to you. We saw it when we started testing more conditions in CI, and hitting Gerrit a lot caused issues. Also in automated deploys to dev/test/production environments in succession would work for dev and test then fail in production, which sucked.

jeffw16:
I think this is a good idea and I think we should do this when we have time. Thank you James!

Ayman161803:
I would like to work on this. Can I please have this assigned?

yaronkoren:
Yes, definitely! Go ahead.

adi-code22:
Is it okay if I work with the same issue?, it seems like a good first issue.

jeffw16:
In the interest of keeping things straightforward, let's keep it one person per task. Please try one of the other tasks suggested in #219. We value completion of harder tasks when selecting the GSoC contributor.

bogisushrith:
Hello @jeffw16 , @yaronkoren , I would like to work on this issue. Could you please assign it to me, I noticed it has been assigned for two weeks.

jeffw16:
@bogisushrith thanks for your interest. This issue already has a code complete PR. It's just awaiting final testing and approval for merging.

yaronkoren:
@bogisushrith (and anyone else) - for those of you who are interested in the Canasta Google Summer of Code project, we have already had interest from a good number of qualified candidates, so we have decided to limit our selection to the group of people who have already submitted a PR in one way or another, to avoid other people wasting their effort. If you want to participate in this year's GSoC, I would say you should find another project, with the Wikimedia Foundation or elsewhere. On the other hand, if you just want to contribute to Canasta, contributions are always welcome!

yaronkoren:
This was done, about six months ago.

PR #220, opened 2023-03-07
chore: switch to github mirrors for cloning

Ayman161803:
Related Issue: #35 

This following PR replaces ``gerrit``  with ``github`` when extensions and skins are cloned

## Areas of discussion
A github mirror for ``SemanticDrilldown`` extension was not found on wikimedia's official github account. However, an alternative was found [here](https://github.com/SemanticMediaWiki/SemanticDrilldown) . The same can replace current SemanticDrillDown gerrit if approved.

## Changes made

Dockerfile gets built successfully after replacing  ``gerrit``  with ``github mirrors``

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,7 +84,7 @@ FROM base as source
 
 # MediaWiki core
 RUN set -x; \
-	git clone --depth 1 -b $MW_CORE_VERSION https://gerrit.wikimedia.org/r/mediawiki/core.git $MW_HOME \
+	git clone --depth 1 -b $MW_CORE_VERSION https://github.com/wikimedia/mediawiki $MW_HOME \
 	&& cd $MW_HOME \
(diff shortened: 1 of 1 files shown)


Ayman161803:
cc @jeffw16  @yaronkoren

jeffw16:
@Ayman161803 thanks for this commit. Successful compilation of the Dockerfile does not necessarily indicate the extensions are all usable. @yaronkoren do you have any good ideas on how to test these 100+ extensions still work as expected?

yaronkoren:
Looks good. I would think just installing Canasta with this Dockerfile, and then doing an "ls -R" within the extensions/ directory would basically do the trick - plus enabling a handful of the extensions to see if they actually work. Presumably, if the same change is done for all of these extensions, then testing a few of them should lead to fairly high confidence that the rest are working as well.

PR #225, opened 2023-03-27
Update VEForAll up to version 0.5

pastakhov:
The CommentStreams extension does not work with 0.4.1 version


yaronkoren:
I agree with this change - actually, there are a number of extensions that I think need to be updated in Canasta in order to work better with MW 1.39, but this is one of them.

jeffw16 (approved):

PR #227, opened 2023-03-30
Fix for SiteNoticeAfter code for nonexistent special pages

yaronkoren:
Going to a nonexistent special page, like Special:Blah, led to an error message instead of the standard display, because of the lack of a "null" check in the Canasta SMTP-handling code.


hexmode (approved):
GH actions still need work, but that would be a separate pull request.

hexmode (approved):

cicalese (approved):

tosfos (approved):

PR #231, opened 2023-04-07
Fix for case when web container can't start after incompletely-shutdown

pastakhov:
The patch to the pid file was not changed when the base image was changed
  from centos to debian.


pastakhov:
It is a major bug. You can get the docker container infinity rebooted with the error message "httpd (pid 369) already running" until the docker container is recreated.

jeffw16 (approved):

PR #248, opened 2023-05-31
Adds SemanticFormsSelect extension

vedmaka:
Adds https://www.mediawiki.org/wiki/Extension:Semantic_Forms_Select version `4.0.0-beta`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -443,6 +443,10 @@ RUN set -x; \
 	&& git clone --single-branch -b $MW_VERSION https://github.com/wikimedia/mediawiki-extensions-SimpleChanges $MW_HOME/extensions/SimpleChanges \
 	&& cd $MW_HOME/extensions/SimpleChanges \
 	&& git checkout -q 5352de89dfaf043f646a44582b26f07822f02be7 \
+	# SemanticFormsSelect
+	&& git clone https://github.com/SemanticMediaWiki/SemanticFormsSelect.git $MW_HOME/extensions/SemanticFormsSelect \
+	&& cd $MW_HOME/extensions/SemanticFormsSelect \
+	&& git checkout 2999c1b3d81e741c843179d1608c1838a5c6eb3a \
 	# SimpleMathJax
 	&& git clone --single-branch https://github.com/jmnote/SimpleMathJax.git $MW_HOME/extensions/SimpleMathJax \
 	&& cd $MW_HOME/extensions/SimpleMathJax \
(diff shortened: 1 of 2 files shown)


yaronkoren:
@vedmaka - what feature(s) of SemanticFormsSelect do you think are useful, that are not already in Page Forms? Ironically, you yourself added some SMW querying syntax to Page Forms a few months ago, which I thought made SFS much less necessary.

vedmaka:
The query feature that was added to PageForms is yet quite limited in comparison to SFS. For example, the feature does not support printouts or any ask params like `mainlabel`. I am not a fan of SFS and, probably, I'd avoid adding SFS, but it's needed by WikiWorks client being migrated to Canasta. Also due to the extension nature (Composer autoload), it can't be mounted through bind on the stack level

yaronkoren:
Well, for non-autocompletion input, like dropdowns, couldn't you just do `values={{#ask:....|format=list}}` ?

Anyway, I think the most persuasive argument for adding SFS to Canasta is that it requires Composer for installation - which means that it would be difficult to add it separately, given the existing issues with Composer add-ons. Although, on that note - I am surprised to see that this current patch adds SFS via both Git and Composer. Are those both necessary?

vedmaka:
> Well, for non-autocompletion input, like dropdowns, couldn't you just do values={{#ask:....|format=list}} ?

I don't think so, SFS allows using dependent fields values in a query runtime, which is not the case for `values={{#ask:` 

> Although, on that note - I am surprised to see that this current patch adds SFS via both Git and Composer. Are those both necessary?

Actually, it's not both, the extension is added via Git, but it still needs `composer.json` to be merged via merge-plugin for autoloading to work. The SFS deps are quite weirdly composed though, so in fact SFS can't be installed via Composer `require`

vedmaka:
I did not use `values dependent on` much, so I am not quite sure how good a replacement for SFS it can be (I guess it still can't be a full replacement because SFS allows using any custom semantic query). Either way, the point is that even if in this particular client case the SFS can be replaced with `values dependent on` it would mean all the client forms to be reworked, which is out of the scope for now.

For the "autoloading" I am referring to the `autoload` section of the SFS `composer.json` https://github.com/SemanticMediaWiki/SemanticFormsSelect/blob/master/composer.json#L53, without inclusion into the merge-plugin the extension simply won't work because necessary classed won't be loaded

yaronkoren:
Well, whatever benefits SFS provides (I still think they are minimal), the fact that it requires Composer for installation makes, in my opinion, a strong reason for including it. Hopefully Composer support will be fixed before the next major Canasta version, and then we can discuss getting rid of SFS (I didn't say get rid of it, I said discuss it), but for now I think it makes sense to include it. So this patch looks good to me.

yaronkoren (approved):

PR #249, opened 2023-06-02
EmbedVideo: use maintained StarCitizenWiki fork

DanielWTQ:
The HydraWiki version is no longer maintained and causes deprecation warnings. Instead of adding patches to fix the issues, switch to an existing and maintained fork of the extension.


yaronkoren:
I think that "403 Forbidden" error that GitHub is issuing is just a weird permissions error that happens when non-"members" submit patches, and is unrelated to the code itself.

tosfos (approved):

jeffw16 (approved):
LGTM

DanielWTQ:
@jeffw16 @tosfos I don't have write access, so I cannot merge this myself

PR #254, opened 2023-06-22
Re-add GTag extension

yaronkoren:
This extension's code has been contained within Canasta since the beginning (it's the one extension that Canasta actually comes bundled with, as opposed to being downloaded via Git or Composer), but it was deleted from the Dockerfile during the temporary "mass deletion" in October 2022:

https://github.com/CanastaWiki/Canasta/commit/3e6d38d401782e0c66cc00d6cb20dc212c70f0ba

It was not re-added, though, when most of the other extensions were re-added two weeks later:

https://github.com/CanastaWiki/Canasta/commit/93e3be6946a90d4d13cddff49594108bdc07998d

I assume that was just a mistake.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -552,6 +552,12 @@ RUN set -x; \
 	&& cd $MW_HOME/extensions/WSOAuth \
 	&& git checkout -q 3c54c4899dd63989bc3214273bf1c5807c7ac5db
 
+# GTag
+COPY _sources/extensions/GTag1.2.0.tar.gz /tmp/
+RUN set -x; \
+    tar -xvf /tmp/GTag*.tar.gz -C $MW_HOME/extensions \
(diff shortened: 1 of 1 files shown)


yaronkoren:
By the way, it turns out that GTag can also be downloaded via Git, from GitHub: https://github.com/SkizNet/mediawiki-GTag/

So there is a strong case to be made for removing the .tar.gz file from Canasta, and switching to a more normal download for this extension. But that is probably best done in another patch.

jeffw16:
How about we ask Skizzerz himself about what we should do in this case?

yaronkoren:
Well, I assume GitHub is the way to go - but I figure we should re-instate the .tar.gz usage for now regardless, so that we have it in the version history.

jeffw16 (approved):

PR #256, opened 2023-06-28
Adds EditAccount extension

vedmaka:


yaronkoren:
There should probably be an issue created for this. But let me ask about this extension - I know very little about it. At the top of the extension's page is a security risk warning:

https://www.mediawiki.org/wiki/Extension:EditAccount

What can you say about the security risk?

tosfos:
It's a bit over the top. The warning is that the extension allows you to change any account's details like email and password, which will allow an admin to take over the account. And that's exactly what the extension is supposed to do. In an enterprise environment I'm not sure this is a significant concern. You can do the same thing with core if you have access to the backend. The warning should probably be changed to say that the extension itself is safe but you should be careful who is granted rights to use it.

yaronkoren:
Okay, that makes sense. (Though I definitely think you should replace that warning!) But please create an issue for this - I don't think it makes sense to have a full discussion in a PR page.

vedmaka:
I have created #259

yaronkoren (approved):

PR #257, opened 2023-06-28
Adds SemanticTasks extension 

vedmaka:
Adds SemanticTasks extension 

Based on #256


yaronkoren:
As with EditAccount, this is better discussed in an issue. Could you please create an issue for this?

vedmaka:
This extension is needed by the WikiWorks client running Canasta, and due to Composer, it can't be mounted on a stack level. Thus I don't think we have many options here. I have created #258

PR #263, opened 2023-07-16
Switches SDU to a forked 1.39 compatible version

vedmaka:
See https://github.com/gesinn-it/SemanticDependencyUpdater/pull/6

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -439,10 +439,11 @@ RUN set -x; \
 	&& git clone --single-branch -b $MW_VERSION https://github.com/wikimedia/mediawiki-extensions-SaveSpinner $MW_HOME/extensions/SaveSpinner \
 	&& cd $MW_HOME/extensions/SaveSpinner \
 	&& git checkout -q 1e819e2fff7fad6999bafe71d866c3af50836c42 \
-	# SemanticDependencyUpdater (v. 2.0.2)
-	&& git clone --single-branch -b master https://github.com/gesinn-it/SemanticDependencyUpdater $MW_HOME/extensions/SemanticDependencyUpdater \
+	# SemanticDependencyUpdater (v. 2.0.2), WikiTeq fork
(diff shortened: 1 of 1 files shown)


yaronkoren:
This seems like a minor bug - why does this need to be fixed at all, in this version of Canasta?

vedmaka:
I can't agree it's a minor bug because the bug effectively makes the extension non-functional on 1.39. Also, there are wikis that are highly dependent on the SDU and this blocks their transition to 1.39 Canasta

yaronkoren:
I don't understand - I thought the use of the "Revision" class was just deprecated, not removed entirely. Is SDU actually broken?

vedmaka:
It was completely removed about a year ago https://gerrit.wikimedia.org/r/c/mediawiki/core/+/684071/ ( https://gerrit.wikimedia.org/r/c/mediawiki/core/+/684075/17 )

yaronkoren:
Wow, it looks like it was removed two years ago! And looking now at its [documentation](https://www.mediawiki.org/wiki/Extension:SemanticDependencyUpdater), I see it clearly states that SDU only supports MediaWiki versions up to 1.36.

I guess that situation with SDU is similar to that with Semantic Forms Select, namely that they:
- should have been tested before getting added to Canasta
- are poorly maintained, to the point perhaps that they should be considered unmaintained
- should be avoided when possible.

That last one, for the case of SDU, seems like it would require switching from Semantic MediaWiki to Cargo - which I think is generally a good idea anyway. (Or using MagicNoCache, or that sort of thing.)

But for this current problem, I'm definitely "warming up" to the idea of switching to a forked version.

gesinn-it-gea:
@yaronkoren SDU and SFS are not unmainained. Well, they could have gotten some more love in the last few months... PRs are welcome.

vedmaka:
Closed in favor of https://github.com/CanastaWiki/Canasta/pull/266

PR #264, opened 2023-07-21
Disable other-vhosts-access-log config for Apache

pastakhov:


vedmaka (approved):
The change looks good, but I'd suggest always try to put at least one line description to PRs

jeffw16 (approved):

Issue #267, opened 2023-07-27
HeaderFooter extension is broken


vedmaka:
Enabling `HeaderFooter` on `1.39` leads to

```
[f5aeea3c86c895618bba39ab] /wiki/Main_Page Error: Call to undefined method ParserOptions::getUser()

Backtrace:

from /var/www/mediawiki/w/canasta-extensions/HeaderFooter/HeaderFooter.class.php(12)
#0 /var/www/mediawiki/w/includes/HookContainer/HookContainer.php(338): HeaderFooter::hOutputPageParserOutput()
#1 /var/www/mediawiki/w/includes/HookContainer/HookContainer.php(137): MediaWiki\HookContainer\HookContainer->callLegacyHook()
#2 /var/www/mediawiki/w/includes/HookContainer/HookRunner.php(2715): MediaWiki\HookContainer\HookContainer->run()
#3 /var/www/mediawiki/w/includes/OutputPage.php(2182): MediaWiki\HookContainer\HookRunner->onOutputPageParserOutput()
#4 /var/www/mediawiki/w/includes/OutputPage.php(2230): OutputPage->addParserOutputMetadata()
#5 /var/www/mediawiki/w/includes/page/Article.php(773): OutputPage->addParserOutput()
#6 /var/www/mediawiki/w/includes/page/Article.php(621): Article->doOutputFromParserCache()
#7 /var/www/mediawiki/w/includes/page/Article.php(528): Article->generateContentOutput()
#8 /var/www/mediawiki/w/includes/actions/ViewAction.php(78): Article->view()
#9 /var/www/mediawiki/w/includes/MediaWiki.php(542): ViewAction->show()
#10 /var/www/mediawiki/w/includes/MediaWiki.php(322): MediaWiki->performAction()
#11 /var/www/mediawiki/w/includes/MediaWiki.php(904): MediaWiki->performRequest()
#12 /var/www/mediawiki/w/includes/MediaWiki.php(562): MediaWiki->main()
#13 /var/www/mediawiki/w/index.php(50): MediaWiki->run()
#14 /var/www/mediawiki/w/index.php(46): wfIndexMain()
#15 {main}
```

jeffw16:
Fixed by #268, closing

PR #268, opened 2023-07-27
Migrates HeaderFooter to Wikia fork

vedmaka:
Enterprisemediawiki version is broken on 1.39, this commit migrates the extension to Wikia version (REL1_39) branch

Solves #267

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -296,9 +296,9 @@ RUN set -x; \
 	&& cd $MW_HOME/extensions/GTag \
 	&& git checkout -q d45f54085d003166aa032363408b8dbef7dd3628 \
 	# HeaderFooter
-	&& git clone https://github.com/enterprisemediawiki/HeaderFooter.git $MW_HOME/extensions/HeaderFooter \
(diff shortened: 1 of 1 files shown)


yaronkoren:
This looks good to me - the official NASA version is clearly not being well-maintained, so I'm glad that at least one organization is maintaining it. (Although I guess it's two, since Hallo Welt has their own version also.) It probably makes sense to talk to the Wikia/Fandom people (and maybe Hallo Welt also) about moving the good version of the code into the Wikimedia repository, but that's a separate conversation.

@vedmaka - one question: the current [Header Footer documentation](https://www.mediawiki.org/wiki/Extension:Header_Footer) says that Composer is the recommended way to install this extension. Is that completely out of date at this point?

vedmaka:
@yaronkoren That's right, in Wikia version it's an outdated instruction (they removed the autoloading part from the extension `composer.json` [[1]](https://github.com/Wikia/HeaderFooter/commit/0f582dd1ddf2d3e79907e064e68534eeff76be90#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34) )

jeffw16 (approved):

PR #269, opened 2023-08-02
Added seqence feature to maintenance script

chl178:


yaronkoren:
@vedmaka - @jeffw16 and I were talking about this PR. Neither of us could remember why these three scripts need to be called in this specific order (if they do). I know that it's been almost a year, but - do you know the reason for this order, if there is one?

vedmaka:
@yaronkoren, I am not sure either. All the maintenance scripts are async by design, so there are no benefits from starting them in order, I guess. Maybe it's good for custom scripts that are supported by that version of the script runner, not really sure

yaronkoren:
@vedmaka - okay, thank you. I'm closing this PR - if at some point later there are scripts added that simply have to be run in a certain order, then presumably they could just be named "mw_1...", "mw_2...", etc.

PR #270, opened 2023-08-04
Allow modification to default composer installed extensions

cicalese:
Move composer installed extensions to volumed in composer.local.json file.
Add support for specifying an alternative composer.local.json file at create/import time.

Depends on CanastaWiki/Canasta-DockerCompose#47
Both pull requests must be merged together.

Optionally extended by CanastaWiki/Canasta-CLI#94, but that pull request does not need to be merged at the same time.

### DIFF
--- a/_sources/configs/composer.canasta.json
+++ b/_sources/configs/composer.canasta.json
@@ -1,21 +1,8 @@
 {
-	"require": {
-		"mediawiki/chameleon-skin": "~4.0",
-		"mediawiki/maps": "10.0.0",
-		"mediawiki/mermaid": "~3.1",
-		"mediawiki/semantic-media-wiki": "4.1.1",
-		"mediawiki/semantic-result-formats": "4.0.2",
-		"mediawiki/semantic-breadcrumb-links": "2.1.x-dev",
-		"mediawiki/semantic-compound-queries": "~2.2",
-		"mediawiki/semantic-extra-special-properties": "3.0.2",
-		"mediawiki/semantic-scribunto": "2.2.0",
-		"mediawiki/simple-batch-upload": "1.8.2",
-		"mediawiki/bootstrap-components": "^5.0",
-		"mediawiki/sub-page-list": "2.0.2"
-	},
 	"extra": {
 		"merge-plugin": {
 			"include": [
+				"/mediawiki/config/composer.local.json",
 				"extensions/AbuseFilter/composer.json",
 				"extensions/CirrusSearch/composer.json",
 				"extensions/DataTransfer/composer.json",


yaronkoren:
Wow, amazing! Now I wonder why adding a user-editable composer.local.json file seemed so difficult to do - apparently it's just a matter of adding one line to the "merge-plugin" section of composer.canasta.json. What was the issue before - do you know?

In any case, I think this is great, though I wonder about the move of all the default extensions (SMW, etc.) from composer.canasta.json to composer.local.json. The idea presumably is to give admins control over whether or not to include them - but putting these extensions (and one skin) in a Composer file only downloads them, it doesn't enable them - no? If so, I would think they should always be downloaded by default - the same as all the Git-based extensions and skins are.

jeffw16:
Thank you Cindy!

Yaron, writing one line of code might take a few seconds to write, but takes a lot longer to think about how to implement it, whether it will actually work, ensure it won't encounter compilation errors, and testing it. Nobody had time to do all of those things until now. It seems stupidly easy, but everyone was busy doing other stuff or didn't understand the promulgated proposal.

jeffw16 (approved):
LGTM

cicalese:
The reason they cannot be downloaded by default is that you may want to use a different version of one of the composer-installed extensions. For git-installed extensions, you have the canasta-extensions directory separate from the user-extensions directory. But, no such distinction is (currently) made for canasta, and any dependencies go in the shared vendor directory. If you want a version of an extension that depends upon different libraries than the default, you could have a conflict. I don't claim that this is a beautiful solution to that problem.

yaronkoren:
Oh... maybe that was the issue; that rings a bell. The ideal solution would have composer.canasta.json place extensions and skins into the canasta-extensions/ and canasta-skins/ directories, and dependencies in vendor/, and would have composer.local.json place extensions and skins into the user-extensions/ and user-skins/ directories, and dependencies in... vendor/ also? But overwriting what was there before? I don't know - that part is indeed trickier. But personally, I'm fine with not guaranteeing that trying to include a different version of one of these 10 or so extensions/skins will work.

Just to be clear - does this current patch lead to all extensions and skins, placed in either composer.canasta.json or composer.local.json, getting installed in the canasta-extensions/ and canasta-skins/ directories? Or I should say, extensions and skins placed in composer.local.json, since under this current patch composer.canasta.json will have none.

cicalese:
> Just to be clear - does this current patch lead to all extensions and skins, placed in either composer.canasta.json or composer.local.json, getting installed in the canasta-extensions/ and canasta-skins/ directories? Or I should say, extensions and skins placed in composer.local.json, since under this current patch composer.canasta.json will have none.

This patch does not affect where the extensions are installed. When using composer/installers, which I'm assuming all of the relevant extensions do, composer will install them in the extensions directory (and skins in the skins directory). I don't know if Canasta moves them from extensions to canasta-extensions after they are installed.

yaronkoren:
Okay, right - I should have thought of that. Well, they all end up in the same directory, anyway. (Or same two directories, one for extensions and one for skins.)

Another question: this might relate more to the corresponding DockerCompose patch, but what exactly happens with the composer.local.default.json file - does it end up on the user's system as composer.local.json?

cicalese:
> Another question: this might relate more to the corresponding DockerCompose patch, but what exactly happens with the composer.local.default.json file - does it end up on the user's system as composer.local.json?

It depends upon whether the user specifies the -c flag to create or install using CLI to provide their own composer.local.json file. If provided, that file (which can be named whatever they want) will be copied to <canasta-instance>/config/composer.local.json, which is accessible from inside the container as /mediawiki/config/composer.local.json. If they do not provide the -c option, composer.local.default.json will be copied to composer.local.json (both residing in <canasta-instance>/config on the host and /mediawiki/config inside the container). In either case, there will be a composer.local.json and a composer.local.default.json file in the config directory, so they can compare their current configuration to the default.

cicalese:
Incidentally, I tested all three patches locally using an override file (using my previous PR) that pointed to a locally built image containing this patch as well as a slightly modified version of the CLI patch that enabled it to pull from a non-main branch of the Canasta-DockerCompose repo (adding `-b <branch>` to https://github.com/CanastaWiki/Canasta-CLI/blob/main/internal/git/git.go#L10).

yaronkoren:
Okay, that's interesting. Sorry I'm asking about all the different patches in one place, but now I'm curious about that new CLI option. How useful is it? If the admin has their own composer.json file (which already, you have to admit, is fairly unlikely), can't they just copy it over after creating the wiki, and restart it? And, for that matter, isn't it better for them to "merge" in whatever Composer lines they have with what is already in the default Composer file, rather than overwriting it?

cicalese:
It is not mandatory, but I do think it is useful. Especially on import, I believe it is beneficial to have the correct extensions in place before running the update script. But, it is definitely a convenience, not essential. It just feels cleaner to me.

yaronkoren:
Well, we can talk about the relative merits of adding a "Composer file" option to the CLI install command, but it doesn't seem like the most important issue for now. Actually - though the three patches are considered a bundle, the CLI patch doesn't have to be merged in at the same time as the other two, does it?

Regardless, I think the extensions and skins being installed via Composer should end up in the user-extensions/ and user-skins/ directories, if that's possible. @jeffw16 (or anyone else) - can it be done?

By the way, arguably this change means that the 10+ extensions/skins installed via Composer will no longer truly be "included in Canasta" but rather some alternate status - "included by default in Canasta", maybe. We can also talk about that separately, though.

jeffw16:
> Regardless, I think the extensions and skins being installed via Composer should end up in the user-extensions/ and user-skins/ directories

How about the bundled extensions using Composer get their dependencies pre-installed in `vendor/` during the image *build* process and those extensions get to remain in `canasta-extensions/`? The user-installed extensions can follow the procedure proposed above.

yaronkoren:
That sounds fantastic, if it's possible... how would that work as far as these extension/skins' placement in any of the Composer JSON files?

cicalese:
> Actually - though the three patches are considered a bundle, the CLI patch doesn't have to be merged in at the same time as the other two, does it?

If it is not merged at the same time, the docker compose patch needs to be modified so that the file retains the name composer.local.json rather than composer.local.default.json.

> Regardless, I think the extensions and skins being installed via Composer should end up in the user-extensions/ and user-skins/ directories, if that's possible. @jeffw16 (or anyone else) - can it be done?

They would have to be installed by composer in extensions and then moved to user-extensions (and the relevant skins directories).

cicalese:
> How about the bundled extensions using Composer get their dependencies pre-installed in `vendor/` during the image _build_ process and those extensions get to remain in `canasta-extensions/`? The user-installed extensions can follow the procedure proposed above.

It would be dangerous to have libraries installed in the vendor directory (and other directories indicated by composer installers) not match the manifest as described in the composer.lock file.

jeffw16:
> It would be dangerous to have libraries installed in the vendor directory (and other directories indicated by composer installers) not match the manifest as described in the composer.lock file.

Oh, so my proposal not allow for the `composer.lock` file to remain consistent with the `vendor/` directory because users could potentially remove the included dependencies from `composer.local.json`? :(

yaronkoren:
I'm still curious to hear the answer to Jeffrey's question, although I'm guessing the answer is "yes".

Anyway - thinking more about the CLI patch and the "composer.local.default.json" setup, I think it needs to change. The issue is that the CLI is meant to be optional - that is, Canasta should be installable without it - but this set of patches makes it mandatory, to some extent. So it seems like, as you suggested, that set of Composer "requires" should be placed in composer.local.json instead of composer.local.default.json. (Though an identical copy of the file could still be placed in composer.local.default.json, if the goal is for admins to always be able to see the original version.)

There's arguably still a problem with this set of patches, which is that make _Docker Compose_ mandatory in a way that it wasn't before, but if Canasta will still work without Docker Compose - other than not having those 10 or so extensions and skins included by default - then I think it's fine.

cicalese:
> I'm still curious to hear the answer to Jeffrey's question, although I'm guessing the answer is "yes".

Yes.

> Anyway - thinking more about the CLI patch and the "composer.local.default.json" setup, I think it needs to change. The issue is that the CLI is meant to be optional - that is, Canasta should be installable without it - but this set of patches makes it mandatory, to some extent. So it seems like, as you suggested, that set of Composer "requires" should be placed in composer.local.json instead of composer.local.default.json. (Though an identical copy of the file could still be placed in composer.local.default.json, if the goal is for admins to always be able to see the original version.)

I think that is a good solution: have the default values in composer.local.json as well as composer.local.default.json. I don't like having two identical files in the repo that could get out of sync. I suggest that the docker compose patch contains only composer.local.json, which would be a simple change to that patch just renaming the file back to its original name. Then, the CLI patch could be changed to make a copy of composer.local.json to composer.local.default.json (better name suggestions absolutely welcome) if it is going to overwrite it.

> There's arguably still a problem with this set of patches, which is that make _Docker Compose_ mandatory in a way that it wasn't before, but if Canasta will still work without Docker Compose - other than not having those 10 or so extensions and skins included by default - then I think it's fine.

There are already other config files that the docker compose repo copies over that would not be there if it is not used, so I do not believe this patch set is creating that problem. I'm actually not a fan of having all of these pieces in separate repos. It creates an illusion of independence that isn't completely accurate while making changes like these much more difficult. I'd rather see a single monorepo with logical components separated into distinct subdirectories. This approach does indeed make it more clear when component boundaries are crossed, however.

yaronkoren:
Those changes sound good. And yes, what you're saying about integrating the Docker Compose (and, presumably, Kubernetes) code into this repository makes a lot of sense. Not for right now, of course!

cicalese:
> Those changes sound good.

Done.

yaronkoren:
Okay, I see. The patches look good to me! Very straightforward, if I understand it correctly: the Composer-loaded extensions were moved from composer.canasta.json to composer.local.json; composer.local.json was "activated" (I guess it wasn't actually usable until now); and there's a new "composer" flag for the CLI create and important commands that's not directly necessary, but could potentially be useful. And perhaps this change means that these ~10 skins and extensions (I'd like to call them the "Jeroen bundle") are no longer quite part of Canasta - they have a different status.

Anyway, this and the Docker Compose patch together are, I think, ready for merging. (The CLI patch should have at least a little more discussion.) Does anyone else have thoughts on these?

cicalese:
> Anyway, this and the Docker Compose patch together are, I think, ready for merging. (The CLI patch should have at least a little more discussion.) Does anyone else have thoughts on these?

I updated the pull request descriptions to reflect that the CLI patch does not need to be merged at the same time.

yaronkoren:
Cool, thank you!

I still hope to hear what anyone else thinks, but if there are no objections, I hope to check these in in the next few days (unless someone else does it first).

yaronkoren (approved):

vedmaka:
@jeffw16 @cicalese @yaronkoren 

I believe we still have a problem here, and a follow-up patch is required:

Composer dependencies are installed during image build time [[1]](https://github.com/CanastaWiki/Canasta/blob/master/Dockerfile#L572), and thus any additions or modifications made to the `composer.local.json` through changing runtime-mounted `/mediawiki/config/composer.local.json` won't have any effect

For the runtime changes to have any effect, it's either necessary to manually run the `composer update` command from within the running container after the file was edited or modify the image runtime init scripts to support this behavior and run the Composer updates automatically. The second is also necessary for the changes to be recovered on container re-creation.

Worth noting that running `composer update` will install skins and extensions into the `skins` and `extensions` directories and generate autoload paths accordingly. This is controlled by composer-installers [[2]](https://github.com/composer/installers) and can be altered, but considering `merge-extra` is disabled by default [[3]](https://github.com/wikimedia/composer-merge-plugin#merge-extra) it requires modifications to the root composer file and be supported by symlinking mechanism.

While altering the composer-installers destination is possible (e.g., set it to `user-extensions` and `user-skins`), the root `vendor` is still a problem for runtime Composer updates because it won't be persisted upon container re-creation and thus also makes it necessary to be handled via runtime init scripts. Otherwise, any dependencies of the extensions installed will be affected too

Yet this set of changes does not contain any modifications to the runtime scripts, and so, in fact, it removes the "Jeroen bundle" from the image and provides no proper way to install this bundle back (same as installing any other Composer dependency). In addition to that, these changes render it impossible to bundle any extension via Composer at build time because autoload is now intended to be regenerated at runtime, making build-time overwrites to the autoload file ephemeral

A possible solution to this could be as below:

* Move the bundled "Jeroen bundle" back to the image Composer file
* Update build-time commands [[1]](https://github.com/CanastaWiki/Canasta/blob/master/Dockerfile#L572) by removing the autoload file rewrites and adding composer-installers alternate path instructions to the root `composer.json` instead
* Modify runtime scripts to perform the `composer update --no-dev` command on start before symlinking
* Modify the symlink watch script to also watch for mounted composer file changes and run `composer update --no-dev`
* Ensure at runtime composer goes before any database update scripts

One of the concerns is that we will have to override the composer-installers install path per package [[4]](https://github.com/composer/installers#custom-install-paths) for the build-time because during build-time we want the extensions to go into `canasta-extensions`, and during runtime, we want them to land into `user-extensions` (though how much sense this makes for composer installed extensions?)

This also leaves a problem related to unwanted runtime updates of `vendor` packages because the merge-plugin ignores `install` command and only re-evaluates the composer files tree on `update` command, which leads to other packages being updated too, and the lock files recalculated. I don't think we have a solution for this though

cicalese:
Agreed the issue of composer support needs more discussion.

I had noticed yesterday that restarting the container does not run composer, although running composer manually did seem to work for me at least with my use case. I'm not sure about `merge-extra` being disabled, but it is possible that was not apparent from my use case.

However, I did notice yesterday that when I tried to use a custom version of SMW installed from git (because I'm testing https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5451), and ran composer manually, composer merged the files listed in https://github.com/CanastaWiki/Canasta/blob/master/_sources/configs/composer.canasta.json#L18, did not detect SMW was already installed since it wasn't in the composer.lock file, and, since SMW is required by Semantic Forms Select and Semantic Tasks, overwrite the version from git with a composer-installed version, which was very undesirable. I'm wondering why those two extensions are listed in the merge section, since they have no dependencies that need to be installed other than SMW in their composer.json file. Removing them would not address the general problem but would at least prevent that behavior.

Perhaps it would be good to set up a time for us all to meet and discuss these issues?

yaronkoren:
Some thoughts:

- I think it's fine if the "Jeroen bundle" does not get included in Canasta by default, and users have to run "composer update", in one way or another, in order to install them - these are now in a separate category from all the other extensions and skins.
- By similar logic, I think it makes sense for these to go into the user-extensions/ and user-skins/ directory.
- I agree that the CLI commands "canasta init", "canasta restart", etc. should also call "composer update", if that's easy to do.
- On the other hand, I'm not sure that there needs to be a "watch" script to re-run "composer update" any time there's a change to composer.local.json. Maybe it's better to just have the user run "canasta restart", or "composer update", when they're ready to make the change official. (Though once Composer does get re-run, hopefully the existing watch script will see the changes to user-extensions/ and user-skins/, and modify the symlinks accordingly.)
- It does seem like Semantic Forms Select and Semantic Tasks should get removed from composer.canasta.json. Will downloading them still work if that is done? Or do _all_ SMW-requiring extensions now need to be moved to composer.local.json, and out of the Dockerfile?

Hopefully we can discuss this all at the next Canasta meeting, if we don't discuss it earlier...

vedmaka:
> and users have to run "composer update", in one way or another

I would be very cautious about this. The fact that there is a need to run `composer update` on a running container is quite a bad thing by itself. This modifies the container state which is against the [statelessness and immutability](https://cloud.google.com/architecture/best-practices-for-operating-containers#ensure_that_your_containers_are_stateless_and_immutable) rule, so unless all the modifications (vendor, composer.lock, all the extension being installed/removed) are persisted via either volumes or bind-mounts this poses a big risk of a deviated state upon stack redeployments, image upgrades and so on

PR #272, opened 2023-08-10
Migrates HeaderFooter to WMF repo version

vedmaka:
BlueSpice has taken over maintenance of the HeaderFooter https://github.com/wikimedia/mediawiki-extensions-HeaderFooter


yaronkoren:
I figured we would do this at some point. Has anyone tested this version of the code to make sure it works and doesn't crash the wiki, etc.?

pastakhov (approved):
I tested it, it works and does not crash the wiki

PR #273, opened 2023-08-13
Adds Flow extension

vedmaka:
Don't merge, please, this conflicts with https://github.com/CanastaWiki/Canasta/pull/270

https://www.mediawiki.org/wiki/Extension:StructuredDiscussions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -279,6 +279,10 @@ RUN set -x; \
 	&& git clone --single-branch -b master https://github.com/wikimedia/mediawiki-extensions-FlexDiagrams $MW_HOME/extensions/FlexDiagrams \
 	&& cd $MW_HOME/extensions/FlexDiagrams \
 	&& git checkout -q eefc9e29aedfc6d8ffaf4f4e50043b390ebd7adc \
+	# Flow
+	&& git clone --single-branch -b $MW_VERSION https://github.com/wikimedia/mediawiki-extensions-Flow.git $MW_HOME/extensions/Flow \
+	&& cd $MW_HOME/extensions/Flow \
+	&& git checkout -q f2998fd1a0676d26c33d97a8272c76fc68b387b6 \
 	# GlobalNotice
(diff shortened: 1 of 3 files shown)


yaronkoren:
Before you do any more work on this patch: I think I can say with pretty high confidence that the Flow/StructuredDiscussions extension will never get added to Canasta. It is barely maintained at this point, and the DiscussionTools extension is a superior solution in just about every way. In fact, I dare say that any Canasta-based wiki that is using StructuredDiscussions should switch off of it, if at all possible. (If there are already many messages within the SD system, that's a problem - hopefully at some point the WMF will create a tool to convert these threads to regular talk pages, but I don't think they have yet.)

vedmaka:
Yes, the [DiscussionTools](https://www.mediawiki.org/wiki/Extension:DiscussionTools) looks nice, and yes, the Flow is used by a WikiWorks customer wiki at the moment ( actually maybe not one but few ), and yes, afaik there is no migration tool available yet. Do we want to be the ones who will create it?

yaronkoren:
That's unfortunate. Yes, perhaps we should be creating such a script, assuming one or more customers would pay for it. But this discussion should probably move elsewhere, since this is no longer a Canasta issue but a WikiWorks issue.

yaronkoren:
Well, since we have been talking about it here: it turns out that StructuredDiscussions already includes a script, [convertToText.php](https://phabricator.wikimedia.org/diffusion/EFLW/browse/master/maintenance/convertToText.php), which takes in the discussion on a single talk page (or any kind of page, really) and converts it all to wikitext. Maybe this is a good enough solution?

yaronkoren (approved):

yaronkoren:
Oh wait, I just meant to approve the "--skip-config-validation" change - I didn't meant to approve this whole PR.

vedmaka:
We should cherry-pick the --skip-config-validation commit

yaronkoren:
Closing this PR, even though it's just a draft. Hopefully this will not cause any problems.

PR #274, opened 2023-08-14
Remove SemanticFormsSelect, Semantic Tasks from composer.canasta.json

yaronkoren:
Neither of these extensions contain any Composer dependencies that need to be downloaded.

### DIFF
--- a/_sources/configs/composer.canasta.json
+++ b/_sources/configs/composer.canasta.json
@@ -14,9 +14,7 @@
 				"extensions/Widgets/composer.json",
 				"extensions/GoogleAnalyticsMetrics/composer.json",
 				"extensions/OpenIDConnect/composer.json",
-				"extensions/WSOAuth/composer.json",
-				"extensions/SemanticFormsSelect/composer.json",
-				"extensions/SemanticTasks/composer.json"
+				"extensions/WSOAuth/composer.json"
 			]
 		}
 	}


vedmaka:
Both extensions are still downloaded via Git [here](https://github.com/CanastaWiki/Canasta/blob/master/Dockerfile#L458) and [here](https://github.com/CanastaWiki/Canasta/blob/master/Dockerfile#L450), and not installed via Composer, thus removing both extensions from the composer file breaks them because the `autoload` sections won't processed anymore [[1]](https://github.com/SemanticMediaWiki/SemanticFormsSelect/blob/master/composer.json#L53) [[2]](https://github.com/SemanticMediaWiki/SemanticTasks/blob/master/composer.json#L20). If we want to keep the extensions, we need to remove both from Dockerfile and instead add into https://github.com/CanastaWiki/Canasta-DockerCompose/blob/main/config/composer.local.json sample file `require` section to allow optional installation

yaronkoren:
Oh, I didn't realize I need to look at the "autoload" section too. I'll revert this change now.

cicalese:
But, they both have autoload statements in their extension.json files. They should not need the autoload statements in the composer.json files unless there is something missing from or incorrect in the statements in extension.json. In fact, it is bad form to have autoloading specified in both places and could potentially result in conflicts and confusion. Is there something wrong with the autoload statements in their extension.json files?

PR #280, opened 2023-08-18
Download all "real" extensions and skins via Git and not Composer

yaronkoren:
The only extensions that are still downloaded via Composer are "helper"/library extensions like DataValues, ParserHooks and Bootstrap (which is to some extent a helper extension).

This PR adds patches for four SMW-based extensions: two (SBL and SRF) get their composer.json requirement for Semantic MediaWiki removed, so that calling "composer install/update" on them does not lead to unnecessary downloads; and the other two (SCQ and SS) get "autoload" added to their extension.json, taking the place of autoload that was called via Composer.

This change should happen concurrently with https://github.com/CanastaWiki/Canasta-DockerCompose/pull/48, which removes all extensions and skins from the file composer.local.json.


Dockerfile
```diff
@@ -439,6 +452,14 @@ RUN set -x; \
 	&& git clone --single-branch -b $MW_VERSION https://github.com/wikimedia/mediawiki-extensions-SaveSpinner $MW_HOME/extensions/SaveSpinner \
 	&& cd $MW_HOME/extensions/SaveSpinner \
 	&& git checkout -q 1e819e2fff7fad6999bafe71d866c3af50836c42 \
+	# SemanticBreadcrumbLinks
+	&& git clone --single-branch -b master https://github.com/SemanticMediaWiki/SemanticBreadcrumbLinks $MW_HOME/extensions/SemanticBreadcrumbLinks \
+	&& cd $MW_HOME/extensions/SemanticBreadcrumbLinks \
+	&& git checkout -q 87a69003743f1de52338f4717cfcf5218ca5a743 \
+	# SemanticCompoundQueries (v. 2.2.0)
+	&& git clone --single-branch -b master https://github.com/SemanticMediaWiki/SemanticCompoundQueries $MW_HOME/extensions/SemanticCompoundQueries \
```
vedmaka, line 460: It seems `SemanticCompoundQueries` is missing from the `composer.canasta.json` file

Dockerfile
```diff
@@ -447,10 +468,22 @@ RUN set -x; \
 	&& git clone --single-branch -b $MW_VERSION https://gerrit.wikimedia.org/r/mediawiki/extensions/SemanticDrilldown $MW_HOME/extensions/SemanticDrilldown \
 	&& cd $MW_HOME/extensions/SemanticDrilldown \
 	&& git checkout -q e960979ec5a3b1e662b3742cee7e7ef4056f9a46 \
+	# SemanticExtraSpecialProperties (v. 3.0.4)
+	&& git clone --single-branch -b master https://github.com/SemanticMediaWiki/SemanticExtraSpecialProperties $MW_HOME/extensions/SemanticExtraSpecialProperties \
```
vedmaka, line 472: `SemanticExtraSpecialProperties` is also missing from the `composer.canasta.json` file

Dockerfile
```diff
@@ -447,10 +468,22 @@ RUN set -x; \
...
+	&& cd $MW_HOME/extensions/SemanticExtraSpecialProperties \
+	&& git checkout -q e449633082a4bf7dcae119b6a6d0bfeec8e3cfe8 \
+	# SemanticScribunto (v. 2.2.0)
+	&& git clone --single-branch -b master https://github.com/SemanticMediaWiki/SemanticScribunto $MW_HOME/extensions/SemanticScribunto \
```
vedmaka, line 476: `SemanticScribunto` is missing from the composer file

vedmaka (changes requested):

Dockerfile
yaronkoren: This extension doesn't need to be added to a Composer JSON file, does it? It doesn't seem like it has any special dependencies.

Dockerfile
yaronkoren: Same question as above.

Dockerfile
```diff
@@ -593,6 +638,30 @@ RUN set -x; \
 	cd $MW_HOME \
 	&& git apply /tmp/core-local-settings-generator.patch
 
+# SemanticBreadcrumbLinks
```
cicalese, line 641: You have to patch the composer.json files before you run composer update.

yaronkoren: Good point! I just fixed this, I think. I also removed the 2nd call to "composer update", which I believe is no longer needed.

_sources/patches/semantic-tasks-composer-reqs.patch
```diff
@@ -0,0 +1,14 @@
...
+index ccad22f..dbb89d1 100644
+--- a/composer.json
++++ b/composer.json
+@@ -9,9 +9,6 @@
+ 	],
+ 	"homepage": "https://www.mediawiki.org/wiki/Extension:Semantic_Tasks",
+ 	"license": "GPL-2.0-or-later",
+-	"require": {
+-		"mediawiki/semantic-media-wiki": "~3.0|~4.0"
+-	},
```
cicalese, line 11: There's nothing left required here. This patch should be removed, and the extension should be removed from composer.local.json.

_sources/patches/semantic-forms-select-composer-reqs.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/composer.json
+@@ -38,9 +38,7 @@
+ 	],
+ 	"require": {
+ 		"php": ">=7.3",
+-		"composer/installers": "1.*,>=1.0.1",
+-		"mediawiki/semantic-media-wiki": "~3.0|~4.0",
+-		"mediawiki/page-forms": ">=5.3.0"
++		"composer/installers": "1.*,>=1.0.1"
+ 	},
```
cicalese, line 11: There are no dependent libraries left here. This patch should be removed, and the extension should be removed from composer.local.json.

_sources/patches/semantic-tasks-composer-reqs.patch
yaronkoren: That's true... both of these extensions are there (and patched) because of the whole "autoload" brouhaha, as discussed at #274. It would be great if @vedmaka could respond to your question there. (Or here, for that matter.) I just don't want to break anything.

_sources/configs/composer.canasta.json
```diff
@@ -4,9 +4,11 @@
 			"include": [
```
cicalese, line 4: If you add:

      "require": {
               "psr/http-message": "1.1"
       },
 
at line 2 of this file, it will prevent an error caused by conflicting versions of this library. As far as I can tell, DataTransfer requires this version, but another extension or library is handled first and pulls version 2.0 instead. Pegging this to version 1.1 works without conflict.

_sources/patches/semantic-tasks-composer-reqs.patch
cicalese: Oh, that's right. I forgot that conversation was still pending. For what its worth, I tested this patch with those two lines removed, and I was able to load both extensions successfully. They both reported they were loaded on Special:Version. But, I don't know what to test to ensure that an error does not occur when the extensions are otherwise exercised.

Dockerfile
yaronkoren: Same question as for the other two.

_sources/patches/semantic-tasks-composer-reqs.patch
yaronkoren: Okay, that's very good to know. I think if they show up on Special:Version, that's pretty much the only test that matters... @vedmaka, I really do hope you respond to this.

_sources/configs/composer.canasta.json
cicalese: DataTransfer requires [phpoffice/phpspreadsheet](https://packagist.org/packages/phpoffice/phpspreadsheet), which requires [psr/http-client](https://packagist.org/packages/psr/http-client), which requires [psr/http-message](https://packagist.org/packages/psr/http-message) version ^1.0 || ^2.0. The error I was getting indicated that by the time DataTransfer was being resolved, http-message version 2.0 had already been selected. It indicated that there http-message 2.0 requires PHP 8, which could not be satisfied, although the spec for http-message says it accepts php: ^7.2 || ^8.0. Basically, it seems that composer got confused, but pegging the version to 1.1 for now fixes the issue.
I got this error when I was running composer update from the command line inside the container after the initial container creation. I had downloaded a patched version of Semantic MediaWiki (to test [this patch](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5451)). This required me to remove the vendor directory (because otherwise I was getting the SemanticBreadcrumbs not an entrypoint error) and run composer update. With the require statement above, all worked perfectly (including running Semantic MediaWiki with the ElasticStore backend).

yaronkoren: Thanks for that detailed explanation - that's very helpful. Now I'm curious about that "vendor" problem - is that something that can be fixed in Canasta? By patching the SemanticBreadcrumbLinks code, possibly?

Also, does this change need to go into this patch, or can it be done separately, i.e. in an earlier patch?

yaronkoren: Man. Another option is to just get rid of SemanticBreadcrumbLinks - it seems like a pretty useless extension, and I don't know why we ever agreed to include it in the first place. (We clearly didn't think it would cause problems!)

I'll create a separate patch, then.

yaronkoren: Okay, done at #282.

Dockerfile
vedmaka: Same thing as with the above, there is an `autoload` section present https://github.com/SemanticMediaWiki/SemanticExtraSpecialProperties/blob/master/composer.json#L55

_sources/patches/semantic-tasks-composer-reqs.patch
vedmaka: Please note that the `require` is not the only reason we load extension via Composer, there is also an `autoload` section which is necessary for PSR autoloading https://github.com/SemanticMediaWiki/SemanticTasks/blob/master/composer.json#L20 https://getcomposer.org/doc/04-schema.md#autoload

_sources/configs/composer.canasta.json
vedmaka: Noting that this was fixed on recent MW versions https://github.com/wikimedia/mediawiki/commit/35cdb41d83c17d56e200824ce56b9d14723586ec

Dockerfile
cicalese: I suggest patching the extension.json file to add an autoload section (and submitting a patch upstream) rather then using composer autoloading for SemanticScribunto.

Dockerfile
cicalese: I suggest patching the extension.json file to add an autoload section (and submitting a patch upstream) rather then using composer autoloading for SemanticCompoundQueries.

_sources/patches/semantic-tasks-composer-reqs.patch
cicalese: But, that autoload section is not necessary for some of the extensions, which also specify autoloading in their extension.json files. See https://github.com/CanastaWiki/Canasta/pull/274#issuecomment-1681466506.

cicalese: @vedmaka Do you agree that if autoloading is done in extension.json rather than composer.json, it is no longer necessary to include the composer.json files for those extensions in composer.canasta.json? Is there any other reason that those composer.json files must be included?

yaronkoren:
The conversation in the code has become chaotic (five separate discussions), so let me try to bring it all inline.

As I understand it, there are five extensions that are being discussed here, because they all include an "autoload" call in their composer.json file, which means that they need to be called with "composer update": Semantic Compound Queries (SCQ), Semantic Extra Special Properties (SESP), Semantic Scribunto (SS), Semantic Tasks (ST) and Semantic Forms Select (SFS). (There are other SMW-based extensions that also include the Composer "autoload", like SRF, SBL and SMW itself, but these need to get a "composer update" regardless because of their dependencies, so it doesn't matter.)

I think the whole "Composer autoload" thing is dumb, but it's not my decision. At this point, from what I understand, we have to deal with it in one of two ways: add all five of these extensions to composer.local.json or composer.canasta.json (ST and SFS are already there), or patch all five extensions to add an "autoload" to their extension.json.

@cicalese and @vedmaka - is that description correct? And if so, any preferences on what to do? My own inclination is to go with Cindy's recommendation to patch all five extensions - the less we have to deal with Composer, the better, I think.

cicalese:
> @cicalese and @vedmaka - is that description correct? And if so, any preferences on what to do? My own inclination is to go with Cindy's recommendation to patch all five extensions - the less we have to deal with Composer, the better, I think.

One correction. Of the five extensions in question, three of them already specify autoloading in their extension.json files, so we can safely remove them from composer.canasta.json and remove any patches created in this PR. Those extensions are 
Semantic Extra Special Properties (SESP), Semantic Tasks (ST) and Semantic Forms Select (SFS).

The other two extensions are missing autoload sections in their extension.json files: Semantic Compound Queries (SCQ) and Semantic Scribunto (SS). Those are the only two that I suggest we patch their extension.json files.

Dockerfile
vedmaka: Good point, that's right! If there is an autoload in the `extension.json`, then the one in the `composer.json` can be ignored

Dockerfile
vedmaka: that's good idea, especially considering there is a `file` autoload which is not cool https://github.com/SemanticMediaWiki/SemanticCompoundQueries/blob/master/composer.json#L44

_sources/patches/semantic-tasks-composer-reqs.patch
vedmaka: Yes, if there is an autoload NS/Classes on the extension scheme (and there are no meaningful `requires` ), it's okay to ignore this section on the composer file

_sources/patches/semantic-forms-select-composer-reqs.patch
vedmaka: That's right, I did miss the autoload on the extension.json

Dockerfile
```diff
@@ -459,6 +492,14 @@ RUN set -x; \
 	&& git clone https://github.com/SemanticMediaWiki/SemanticFormsSelect.git $MW_HOME/extensions/SemanticFormsSelect \
 	&& cd $MW_HOME/extensions/SemanticFormsSelect \
 	&& git checkout 4b56baa752401b4ff9fe555fd57fc5c3309601d4 \
```
cicalese, line 494: add -q flag

yaronkoren: Right, I noticed that before. I think the "-q" is not strictly necessary, although it's good at least for consistency. This fix probably belongs in a separate patch, though.

cicalese: True, this line was not introduced here.

vedmaka:
> @cicalese and @vedmaka - is that description correct? 

yep (+taking into account @cicalese comment)

> And if so, any preferences on what to do? My own inclination is to go with Cindy's recommendation to patch all five extensions

That's right, in this situation, we either patch the extensions `extension.json` or the `composer.json`

> the less we have to deal with Composer, the better, I think.

I am unsure if dealing less with Composer is the right thing to do in the long term. The discussion about using Composer as a main extensions management tool [[1]](https://www.mediawiki.org/wiki/Composer/For_extensions) is still in progress [here](https://phabricator.wikimedia.org/T250406). It could be that managing all the extensions via composer (even non-Composer ones) is a good idea, especially considering the `update` problem is somewhat solved here https://github.com/wikimedia/composer-merge-plugin/commit/1b1afcb69fe9fae36eec90d4f3c865e0e79dfff1 and here https://github.com/wikimedia/composer-merge-plugin/pull/257

In theory, this provides a good number of benefits, like having a parsable composer file on the image instead of tons of `git clone ..` commands and a free version and extensions dependencies checks on extensions being installed

yaronkoren:
Thanks to both of you for your comments. @vedmaka - you may be right that, in the future, a pure Composer-based approach will be the way to go. For now, though, this kind of hybrid approach seems to be causing unnecessary complexity and difficulty, both for admins and for us developers.

I have been making changes to the code and to this PR, based on the recommendations. I think the big thing left to do now is to create patches for SCQ and SS. I'm not sure what to do add to extension.json for either extension - can anyone offer a hint?

cicalese:
> I have been making changes to the code and to this PR, based on the recommendations. I think the big thing left to do now is to create patches for SCQ and SS. I'm not sure what to do add to extension.json for either extension - can anyone offer a hint?

I was going to work on patches for those two, but I'm currently having trouble running with this patch. I see that there is a merge conflict in the composer.canasta.json file that needs to be resolved. Could you please fix that? The error I'm getting is that it is missing Vector, though, so I'm not sure that's related. I'll let you know if I determine there's another issue.

yaronkoren:
Oh yeah, I should have noticed that merge conflict. I just fixed it, I think.

cicalese:
> I think the big thing left to do now is to create patches for SCQ and SS. I'm not sure what to do add to extension.json for either extension - can anyone offer a hint?

For SemanticCompoundQueries/extension.json:
- replace:
	"load_composer_autoloader":true,
- with:
	"AutoloadNamespaces": {
		"SCQ\\": "src/"
	},
	"AutoloadClasses": {
		"SemanticCompoundQueries": "SemanticCompoundQueries.php"
	},

For SemanticScribunto/extension.json:
- replace:
	"load_composer_autoloader":true,
- with:
	"AutoloadNamespaces": {
		"SMW\\Scribunto\\": "src/"
	},
	"AutoloadClasses": {
		"SemanticScribunto": "SemanticScribunto.php"
	},

I successfully tested these modifications on my local instance.

yaronkoren:
Okay, thanks, that's great! Are you sure that the "load_composer_autoloader" lines to be removed, though? They're there for some of the other extensions:

https://github.com/SemanticMediaWiki/SemanticTasks/blob/master/extension.json#L35
https://github.com/SemanticMediaWiki/SemanticFormsSelect/blob/master/extension.json#L66

cicalese:
> Okay, thanks, that's great! Are you sure that the "load_composer_autoloader" lines to be removed, though? They're there for some of the other extensions:
> 
> https://github.com/SemanticMediaWiki/SemanticTasks/blob/master/extension.json#L35 https://github.com/SemanticMediaWiki/SemanticFormsSelect/blob/master/extension.json#L66

Yes. See https://www.mediawiki.org/wiki/Manual:Extension.json/Schema#load_composer_autoloader. Since those extensions do not have any dependencies on libraries that are specified in composer.json, that line can/should be removed. It should actually be removed from the other extensions you referenced as well, but that's a problem for another day.

vedmaka:
By the way, I believe the [load_composer_autoloader](https://www.mediawiki.org/wiki/Manual:Extension.json/Schema/en#load_composer_autoloader) documentation is not well worded. This flag forces local `./vendor/autoload.php` from the **extension directory** to be loaded. This has no use when the merge-plugin is in place because when the extension Composer file is loaded through the merge-plugin all the dependencies are added to the main autoload file and no `vendor` is present at the extension directory

Dockerfile
yaronkoren: I added a patch for SCQ.

yaronkoren:
I just added the autoload patches for SCQ and SS to this PR - including the "load_composer_autoloader" removal. If it turns out that this line is harmless, then it will be easy to, er, remove that removal. In any case, assuming this all works, it will be good to at least add those new autoload lines "upstream" to the SCQ and SS extensions.

cicalese:
Tested. LGTM.

cicalese (approved):

vedmaka (approved):

PR #282, opened 2023-08-24
Add psr/http-message requirement to composer.canasta.json

yaronkoren:
To ensure that the best version of this library is downloaded, given conflicting requirements by different extensions. As suggested by @cicalese.


cicalese (approved):
LGTM

PR #283, opened 2023-08-24
Re-remove SemanticFormsSelect, Semantic Tasks from composer.canasta.json

yaronkoren:
It turns out that their Composer "autoload" is not needed either, because their extension.json files already contain an autoload.


cicalese (approved):
LGTM

PR #285, opened 2023-08-28
Update base to debian 11.7

bawolff:
Use the more up to date 11.7 as the base.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:11.5 as base
+FROM debian:11.7 as base
 
 LABEL maintainers=""
 LABEL org.opencontainers.image.source=https://github.com/CanastaWiki/Canasta


yaronkoren:
Generally, the policy is only to do software updates alongside MediaWiki releases, for the sake of reliability/avoiding surprises - unless there's a bug that needs fixing, in which case of course the rules are different. Is there a specific reason to upgrade to Debian 11.7?

tosfos:
It's due to vulnerabilities in 11.5. See [here](https://hub.docker.com/layers/library/debian/11.5/images/sha256-a4b14cf13428c71a5fce6d9380eb2ac1452a7ae4253e594dc90c8a884fdf3db2?context=explore) .

jeffw16 (approved):

yaronkoren:
I'm not sure what the right answer is about updating Debian if there are no known issues, but a security vulnerability is definitely a good reason to upgrade.

tosfos:
Thanks guys!

PR #299, opened 2023-09-22
Adds SemanticWatchlist v 1.3.0

vedmaka:
Adds https://github.com/SemanticMediaWiki/SemanticWatchlist extension, version 1.3.0

METHW-151


yaronkoren (approved):

yaronkoren:
This is not the "correct" way to add extensions to Canasta (an issue should be created first, and then others have to agree to it), but since this seems like an uncontroversial addition, I'm just approving it anyway.

PR #306, opened 2023-09-24
Change "chameleon" skin directory to be all lowercase

### LINKED ISSUE #304: Chameleon skin incorrect directory name
The Chameleon skin directory name should be lowercase as per https://github.com/ProfessionalWiki/chameleon/blob/master/docs/installation.md (and probably for historical reasons)

The regression originated from https://github.com/CanastaWiki/Canasta/pull/280

yaronkoren:
Fixes #304.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,8 +101,8 @@ RUN set -x; \
 RUN set -x; \
 	cd $MW_HOME/skins \
  	# Chameleon (v. 4.2.1)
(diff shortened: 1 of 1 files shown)


vedmaka (changes requested):
I think we also need to update the https://github.com/CanastaWiki/Canasta/blob/master/_sources/configs/composer.canasta.json file

vedmaka (approved):

PR #312, opened 2023-10-14
Adds MassMessageEmail to composer file

vedmaka:


yaronkoren:
It does indeed look, based on MassMessageEmail's composer.json file, that it has two required libraries: html2text and ext-xml. On ther other hand, the extension's documentation page (https://www.mediawiki.org/wiki/Extension:MassMessageEmail) doesn't say anything about these two. Does the documentation just need to be updated?

vedmaka:
That's right, the doc was updated https://www.mediawiki.org/w/index.php?title=Extension%3AMassMessageEmail&diff=6178232&oldid=5987703

yaronkoren:
Great! Everything looks good now.

yaronkoren (approved):

PR #347, opened 2024-01-04
Updated to use debian 12.4 and PHP 8.2

cicalese:
This PR updates debian from 11.7 to 12.4 and PHP from 7.4 to 8.2.

It is currently premature to merge this, since it will likely break support for Semantic MediaWiki and possibly other extensions.

Note that the change from using environment variables to using literals in _sources/configs/php-fpm-www.conf was necessary to prevent errors for some reason. It isn't clear why environment variables no longer work or whether something else was going on.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:11.7 as base
+FROM debian:12.4 as base
 
 LABEL maintainers=""
 LABEL org.opencontainers.image.source=https://github.com/CanastaWiki/Canasta
@@ -47,21 +47,21 @@ RUN set x; \
 	poppler-utils \
 	&& aptitude update \
 	&& aptitude install -y \
-	php7.4 \
-	php7.4-mysql \
-	php7.4-cli \
-	php7.4-gd \
-	php7.4-mbstring \
-	php7.4-xml \
-	php7.4-mysql \
-	php7.4-intl \
-	php7.4-opcache \
-	php7.4-apcu \
-	php7.4-redis \
-	php7.4-curl \
-	php7.4-zip \
-	php7.4-fpm \
-	php7.4-yaml \
+	php \
+	php-mysql \
+	php-cli \
+	php-gd \
+	php-mbstring \
+	php-xml \
+	php-mysql \
+	php-intl \
+	php-opcache \
+	php-apcu \
+	php-redis \
+	php-curl \
+	php-zip \
+	php-fpm \
+	php-yaml \
 	libapache2-mod-fcgid \
 	&& aptitude clean \
 	&& rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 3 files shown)


cicalese:
> I imagine changing to Debian 12.4 in and of itself does not break SMW. But I'm guessing the upgrade to PHP 8.2 was necessary for other reasons. Is PHP 7.4 not supported on Debian?

Agreed that the change in debian is unlikely to break SMW. PHP 8.2 is the only supported PHP on debian 12.4 without adding other repos. And, PHP 7.4 has been end of life with no security fixes for over a year. In fact, PHP 8.0 is also end of life, and PHP 8.1 is only getting security fixes.

I'm curious why the Docker build and push job was run, failing (appropriately) when it attempted to push the container. Isn't it premature to push the container? Maybe I don't understand what that is supposed to be doing.

tosfos:
Thanks for this update! My understanding is that PHP 8.2 has only experimental support in MediaWiki 1.39.

jeffw16:
The CI/CD pipeline creates a build for every push (provided no weird edge case occurs) for testing purposes. The failure is most likely because @cicalese does not have push permissions for the Canasta GHCR.

cicalese:
I'm going to create another pull request with PHP 8.1 support. There are several possible approaches. I could add a repo to pull PHP 8.1 from to an image build from debian:12.4, as this patch is. Or I could build on php:8.1, which also builds on debian:12.4. Or even php:8.1-apache. Note that mediawiki:1.39 (the official MediaWiki image on dockerhub) is built on php:8.1-apache. It could simplify the configuration considerably to build on one of those images. Is there a reason they are not used?

yaronkoren:
If you're asking why we went with PHP 7 and not PHP 8 for the 1.39 version of Canasta - I don't think it was ever really discussed, but I think there was a general sense that the extensions were not uniformly ready for the jump. (Which apparently is still the case now!) But I would think it's pretty much guaranteed that the next major release of Canasta will include PHP 8, ideally the same exact version that the main MediaWiki image will use.

cicalese:
Actually, I was asking why the image was built on debian:X, rather than php:X or php:X-apache (or mediawiki:1.39 even, but I think that's a slightly different discussion).

yaronkoren:
Oh.🤦That I don't know! If there's an image like php:X-apache that includes PHP, Debian and Apache, that does seem like it would simplify things quite a bit. My guess is that just no one thought of doing that.

cicalese:
OK, cool. I'll try a patch built on php:8.1-apache.

jeffw16:
Oh wait, I do know the reason behind this! The WMF runs MediaWiki production on Debian, so we wanted to keep everything on Debian. Plus, we're already locked into using Aptitude instead of Yum. The thought in 2022 was, we would rather have fine-grained control over which OS, PHP, and web server we use in case we want to jump ship to using something different in the future.

cicalese:
But php:8.1-apache is built on debian, so that would satisfy that constraint.

jeffw16:
That's awesome to hear! Is there any risk they might switch to another OS in the future?

cicalese:
Highly unlikely.

yaronkoren:
Closing this, since we're probably going to upgrade to PHP 8.1 instead, and probably in a different way from how this PR does it. The new issue for this change is #352.

PR #358, opened 2024-02-16
Rename run-apache.sh to run-all.sh

### LINKED ISSUE #217: Rename, and possibly simplify, run-apache.sh
Right now the run-apache.sh script, contrary to its rather specific name, does a bunch of things:

1. Creates symlinks for all the extensions and skins
2. Syncs all the files, using rsync
3. Calls update-docker-gateway.sh
4. Checks permissions for the relevant directories, and modifies them if they were not set correctly
5. Keeps calling wait-for-it.sh until the wiki DB is generated
6. Exits if the /user-extensions and /user-skins directories are not set up correctly (this should have been done first, no?)
7. Calls MW's update.php (but only if LocalSettings.php exists, for some reason)
8. Calls mwjobrunner.sh (which in turn calls MW's runJobs.php for the main job types, like "refreshLinks")
9. Calls mwtranscoder.sh (which in turn also calls runJobs.php, but for "webVideoTranscode" jobs)
10. Calls mwsitemapgen.sh (which in turn calls MW's generateSitemap.php)
11. Finally, runs Apache 😄 

At the very least, I think this script should be renamed to something indicating that it's a sort of master do-everything kind of script: setup-all.sh, setup.sh, something like that. It also could be good if any of steps 1, 2 and 4 could be turned into their own scripts, to try to break up the code more. Step 1 takes about 40 lines, step 2 takes about 10 lines, and step 4 takes about 20 lines; so step 1 would be the most obvious candidate for splitting off, into a script called create-symlinks.sh or something.

yaronkoren:
Fixes #217 .


jeffw16:
This doesn't just set things up. It also runs everything over and over again; it's the "main thread" of the container after startup. I strongly prefer a different name.

Dockerfile
```diff
@@ -783,4 +783,4 @@ WORKDIR $MW_HOME
 HEALTHCHECK --interval=1m --timeout=10s \
 	CMD wget -q --method=HEAD localhost/w/api.php
 
-CMD ["/run-apache.sh"]
+CMD ["/setup-all.sh"]
```
jeffw16, line 786: This is a misnomer too.

jeffw16 (changes requested):

yaronkoren:
What name would you suggest?

yaronkoren:
@jeffw16 (or anyone else) - what about run-all.sh?

jeffw16:
Honestly I think `run.sh` would be fine. `run-all.sh` is alright too.

yaronkoren:
Alright, it's now being renamed to run-all.sh. I like the new name - it reminds me of the renaming of the VEForPageForm extension to "VEForAll".

jeffw16 (approved):

PR #361, opened 2024-02-27
Upgrade PHP7.4 to PHP8.1

### LINKED ISSUE #352: Upgrade Canasta to PHP 8.1
Canasta currently runs with PHP 7.4, which unfortunately has been end-of-life since November 2022, meaning that it no longer accepts security fixes. That's a big problem. Thus, we should upgrade to PHP 8.1 as soon as possible.

PHP 8.1 will itself be EOL in November 2024, but unfortunately that's the best we can do, because MW 1.39 does not yet officially support PHP 8.2 or higher. In fact, even the very latest MW version, 1.42, doesn't yet support PHP 8.2! You can see the Phabricator workboard to complete PHP 8.2 support [here](https://phabricator.wikimedia.org/project/view/6053/) - hopefully this will be done by the end of the year.

As discussed at #347, it appears that the ideal approach for this is to change Canasta to use php:8.1-apache as its base image, instead of debian:11.7. Since php:8.1-apache includes PHP, Apache and Debian, this change should allow for the Dockerfile to be simplified quite a bit. Hopefully this won't cause any problems with PHP-FPM or anything else.

A note about extensions and skins: there are some (like Page Forms and Cargo) that will need to be updated, because their latest versions have improved support for PHP 8. There may be others (possibly Semantic MediaWiki) that still don't fully work with PHP 8. For those, there's nothing much we can do - assuming they work at all, just include them and hope their remaining bugs are not too severe.

TechieNK:
Fixes #352 .

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,7 @@ RUN set x; \
 	apt-transport-https \
 	ca-certificates \
 	wget \
+	lsb-release \
 	imagemagick  \
 	librsvg2-bin \
 	python3-pygments \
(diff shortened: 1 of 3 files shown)


jeffw16:
Has this been tested yet?

yaronkoren:
@TechieNK tested it, and it worked for him - I think he's the only one that has tested it.

I should note that this is not a complete patch yet - I went through the Canasta extensions, and at this point I know of four that need their version upgraded in the Dockerfile to work better with PHP 8: Cargo, Page Forms, Semantic MediaWiki and UserPageViewTracker.

jeffw16:
Let's hold off on merging this until compatibility is confirmed with those extensions.

yaronkoren:
Moving this to the local "php-8.1" branch.

Issue #321, opened 2023-11-14
The create-symlinks.sh script does not remove existing links in extensions folder 


pastakhov:
How to reproduce: restart a web docker container without recreating one. You will see the error messages like:
```
Symlinking bundled extensions...
ln: failed to create symbolic link '/var/www/mediawiki/w/extensions/AJAXPoll/AJAXPoll': File exists
...
```
For each extension.

Most likely, it is the reason for #300 

When you remove an extension from a folder mounted to the `user-extensions` folder and restart the container you expect that the symlink in the extension folder will be changed to use the Canasta bundled extension. However because the script does not delete existing symbolic links, the symlink to the removed extension from the `user-extensions` folder will not be replaced.

pastakhov:
I agree, the script is not working, but I didn't test/check the monitor-directories.sh script. I thought it was removed or disabled. It can be related to this issue (or be similar), but it is another one. This issue describes a mistake in the create-symlinks.sh script.  In other words, the monitor-directories.sh does not affect this issue but can have similar mistake in the code (and yes, it does not work).

yaronkoren:
That's true - the problem needs to be fixed in create-symlinks.sh. I still think fixing monitor-directories.sh (or maybe just getting it to always run) will fix this problem 99% of the time, but there will still be cases when a "bad" symlink is there when Canasta is started, and for that, create-symlinks.sh needs to take action.

yaronkoren:
I believe this was fixed by PR #366 ... I'm closing this now.

PR #366, opened 2024-03-06
Fix monitor-directories script and create-symlinks script

### LINKED ISSUE #363: The inotifywait function call was removed and /monitor-directories.sh is not running
This commit https://github.com/CanastaWiki/Canasta/commit/f962d3db75f045fe5907ad1758ddc66525842ff9 removed (most likely by mistake) the line with `inotifywait &`  and  /monitor-directories.sh is not running in the docker container. When you add/remove an extension in the `user-extensions` folder, the symlinks are not updated.

TechieNK:
Fixes #363 and #321 .


yaronkoren (approved):
This looks great!

_sources/scripts/run-apache.sh
```diff
@@ -181,10 +173,6 @@ check_mount_points () {
   fi
 }
 
-inotifywait() {
-	/monitor-directories.sh
```
jeffw16, line 185: Why isn't this file deleted? Also does @vedmaka approve of this removal of inotifywait?

_sources/scripts/create-symlinks.sh
```diff
@@ -19,21 +19,13 @@ for user_extension_path in $(find $MW_HOME/user-extensions/ -maxdepth 1 -mindept
 do
   user_extension_id=$(basename $user_extension_path)
   extension_symlink_path="$MW_HOME/extensions/$user_extension_id"
-  if [[ -e "$extension_symlink_path" ]]
-  then
-    rm "$extension_symlink_path"
-  fi
-  ln -s $MW_HOME/user-extensions/$user_extension_id/ $MW_HOME/extensions/$user_extension_id
+  ln -sf $MW_HOME/user-extensions/$user_extension_id/ $MW_HOME/extensions/$user_extension_id
```
jeffw16, line 22: How do we know for sure this has different behavior from the statements it replaced?

jeffw16 (changes requested):

_sources/scripts/run-apache.sh
yaronkoren: To be clear, the monitor-directories.sh script is still geting called - or rather, it was (accidentally) not getting called after a previous patch, but with this PR it's getting called again - this time by including it in the maintenance-scripts/ folder.

_sources/scripts/create-symlinks.sh
yaronkoren: @TechieNK tried it, and it works for him. It makes sense - adding the "-f" flag means that ln automatically deletes whatever symlink was there before, which is what that "if" statement does also.

yaronkoren (approved):

_sources/scripts/create-symlinks.sh
jeffw16: It'd be great to get an explanation. Anyone who wants their code merged in should be able to justify all aspects of their code. "It works for me" isn't enough for me to approve a PR because it shows there's still uncertainty about what this does.

yaronkoren: It's basically explained in the commit message, no? There's the item "Replace symlink delete with force write." What more needs to be said about this particular code change?

jeffw16: I don't understand what the difference between deletion/creation and force write is.

jeffw16: Also, we should add some comments here explaining why we are doing this. Comments are awesome (contrary to what people at Google might say).

yaronkoren: Do you mean comments in the code, or comments in the commit message? It sounds like the former, but I don't know what it means to explain a _change_, within the code.

_sources/scripts/run-apache.sh
```diff
@@ -208,7 +196,14 @@ fi
 echo "Starting services..."
 
 run_maintenance_scripts &
-inotifywait &
+
+echo "Checking permissions of $MW_VOLUME/sitemap..."
```
pastakhov, line 200: Why do you think it should check permissions after you run the maintenance scripts?

_sources/scripts/create-symlinks.sh
```diff
@@ -19,21 +19,13 @@ for user_extension_path in $(find $MW_HOME/user-extensions/ -maxdepth 1 -mindept
 do
   user_extension_id=$(basename $user_extension_path)
   extension_symlink_path="$MW_HOME/extensions/$user_extension_id"
```
pastakhov, line 21: Where is the `extension_symlink_path` variable used?

_sources/scripts/create-symlinks.sh
```diff
@@ -19,21 +19,13 @@ for user_extension_path in $(find $MW_HOME/user-extensions/ -maxdepth 1 -mindept
...
-  fi
-  ln -s $MW_HOME/user-extensions/$user_extension_id/ $MW_HOME/extensions/$user_extension_id
+  ln -sf $MW_HOME/user-extensions/$user_extension_id/ $MW_HOME/extensions/$user_extension_id
 done
 
 echo "Symlinking user skins and overwriting any redundant bundled skins..."
 for user_skin_path in $(find $MW_HOME/user-skins/ -maxdepth 1 -mindepth 1 -type d)
 do
   user_skin_id=$(basename $user_skin_path)
   skin_symlink_path="$MW_HOME/skins/$user_skin_id"
```
pastakhov, line 29: Where is the `skin_symlink_path` variable used?

_sources/scripts/create-symlinks.sh
jeffw16: Just add the comment: `# -f flag works better than removing the symlink and freshly linking it because ...`

yaronkoren: Isn't it obvious why a single command is better than multiple commands that do the same thing?

jeffw16: But they seem to do different things entirely, so I really would like some info in the code about why it's better.

jeffw16: I was just informed that they don't do anything different and this didn't fix the bug by itself. Okay, then my objection is lifted.

jeffw16 (approved):

PR #372, opened 2024-03-14
Upgrade PHP7.4 to PHP8.1

### LINKED ISSUE #352: Upgrade Canasta to PHP 8.1

yaronkoren:
Fixes #352. Patch mostly by @TechieNK .

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -432,7 +432,7 @@ RUN set -x; \
 	&& git clone --single-branch -b $MW_VERSION https://github.com/wikimedia/mediawiki-extensions-PageExchange $MW_HOME/extensions/PageExchange \
 	&& cd $MW_HOME/extensions/PageExchange \
 	&& git checkout -q 28482410564e38d2b97ab7321e99c4281c6e5877 \
-	# PageForms (v. 5.7.1)
+	# PageForms (v. 5.7.2)
 	&& git clone --single-branch -b master https://github.com/wikimedia/mediawiki-extensions-PageForms $MW_HOME/extensions/PageForms \
 	&& cd $MW_HOME/extensions/PageForms \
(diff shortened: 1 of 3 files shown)


yaronkoren:
The Dockerfile now includes upgrades to the four extensions that I'm aware have had PHP 8 fixes since the last Canasta release: Cargo, Page Forms, Semantic MediaWiki and UserPageViewTracker.

yaronkoren:
Replaced by #379.

PR #374, opened 2024-03-21
Further fix for Maps with VE

yaronkoren:
Another fix by @pastakhov. I don't know if this one is as critical as the first one.


pastakhov:
it is critical

pastakhov (approved):
I tested the updated extension

PR #376, opened 2024-03-23
Move extensions setup to extensions.yaml and add python script for execution

### LINKED ISSUE #291: Move list of Git-downloaded extensions into a YAML file
About 500 of the Canasta Dockerfile's 761 lines are devoted to one thing: downloading the roughly 120 extensions that Canasta gets via Git. The code for every download is roughly the same, so this represents a lot of duplicated code. This could be done much more cleanly by putting all the needed information about these extensions into a single file (probably YAML makes the most sense), then having the Dockerfile read this file and cycle through the extensions, downloading each one. Having this separate file could also make it easier for people to use Canasta as a base image for their own MediaWiki distributions, since presumably the big difference between one distribution and the next is the set of extensions used.
	
(Very similar code could be created for downloading skins - this issue covers only extensions, since the need for streamlining skin downloads seems much smaller. But it would be great to do both, I think.)

TechieNK:
Fixes #291 .


jeffw16:
Why are we using Python all of a sudden? Please use a compiled language rather than an interpreted language.

jeffw16 (changes requested):
I am not approving this until it's rewritten in Bash or a compiled language.

jeffw16:
I do not support adding unnecessary dependencies. As President Kennedy once said, we don't do things because it's easy, but because it's hard. I am sure Python seems like the obvious choice for ease of YAML parsing, but if you look to `jq` in Bash you'll see it's actually quite possible to do it in lower level languages.

yaronkoren:
@jeffw16 - thanks for looking at this. I have no idea what that quote means in this context, and I don't think the speed difference matters, but I do agree that it would be better to have this as a shell script, if it's possible. Hopefully those libraries you mentioned make it easy to do.

_sources/scripts/extension-setup.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+MW_HOME="$MW_HOME"
+MW_VERSION="$MW_VERSION"
+commands=$(yq eval '. | to_json' extensions.yaml)
+echo "$commands" | jq -r '.extensions' | jq -c '.[]' | while read -r obj; do
+    extension_name=$(echo "$obj" | jq -r 'keys_unsorted[]')
```
jeffw16, line 7: Seems a bit repetitive to echo the same object four times and call jq on it. Any way to assign this into a variable or is that not necessary in your opinion, and if not, why not?

_sources/scripts/extension-setup.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
+
+MW_HOME="$MW_HOME"
+MW_VERSION="$MW_VERSION"
+commands=$(yq eval '. | to_json' extensions.yaml)
```
jeffw16, line 5: Any reason we are using yq and then converting to jq? Please add a comment explaining why, since this is not immediately obvioua.

jeffw16 (changes requested):
Super glad to see this rewritten in Bash. I only have two small requests for you

_sources/scripts/extension-setup.sh
TechieNK: echo and using it as a variable both have the same time complexity in my opinion. Even in terms of space, echo won't consume much as we are already accessing the actual data using $obj object

jeffw16: Okay, but this isn't an algorithm, this is a program. It doesn't matter if the time complexity is the same (which it's not, by the way) and space complexity is not an issue here because we are not operating in a low memory environment.

Please change it.

TechieNK: jq command can be applied only on an active stream like echo output. If we store that in a variable, we will again be echoing out and that won't be any better than the current code.

jeffw16: GitHub Copilot says:

No, calling echo and jq -r four times in succession is not the most efficient approach. Each call to echo and jq -r spawns a new process, which adds overhead due to the need for process creation and termination, and inter-process communication.

In the provided script, jq -r is used to extract data from a JSON object. This operation is repeated for extension_name, repository, commit, and branch, which can be optimized.

A more efficient approach could be to parse the JSON object once and assign the results to variables in a single operation. Here's an example of how you could do this:

```
read -r extension_name repository commit branch << EOF
$(echo "$obj" | jq -r '[.extension_name, .repository, .commit, .branch] | @tsv')
EOF
```

This uses jq to create a tab-separated string of the required values, which is then read into the variables in one step. This reduces the number of processes created and should improve the efficiency of the script.

image

jeffw16 (changes requested):

jeffw16 (approved):
Wonderful work, thank you! I hope these various iterations were not too irksome for you. I'm also hopeful this was a good learning opportunity, as writing good code here will pay dividends later if you work for a prestigious tech company (and of course for future open source projects too!).

TechieNK:
> Wonderful work, thank you! I hope these various iterations were not too irksome for you. I'm also hopeful this was a good learning opportunity, as writing good code here will pay dividends later if you work for a prestigious tech company (and of course for future open source projects too!).

Thank you. I did learn about yq and jq :)

PR #390, opened 2024-04-20
Add PHP script for extensions and skins setup

### LINKED ISSUE #401: Use of AMD-specific yq binary causes extension loading to fail in ARM architecture
I believe this AMD-specific line causes Canasta setup to fail on any computer that uses ARM architecture, which is most notably Apple computers (?):

https://github.com/CanastaWiki/Canasta/blob/master/Dockerfile#L141

Perhaps the best solution is to use a YAML parser that is not architecture-specific? I don't know much about YAML parsers, but I would have to imagine that there's some YAML-parsing library or script that will work across all machines... I mean, PHP even contains a yaml_parse() function, so I would think there are any number of ways to do it.

naresh-kumar-babu:
Fixes #401 (Issue became obsolete since we're switching to PHP)

It is much easier to do parsing of YAML in PHP than in .sh; the standard way to do YAML parsing in a shell script is to call yq, but yq requires different executables for different architectures. Given that the YAML parsing is already in PHP, it seemed easier to do the whole thing in PHP rather than split the work between two scripts.


Dockerfile
```diff
@@ -103,30 +103,17 @@ RUN set -x; \
 # Skins
 # The Minerva Neue, MonoBook, Timeless, Vector and Vector 2022 skins are bundled into MediaWiki and do not need to be
 # separately installed.
+COPY _sources/scripts/extensions-skins.sh /tmp/extensions-skins.sh
+COPY _sources/patches/* /tmp/
+RUN wget -qO /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64
```
vedmaka, line 108: this should have a conditional load for arm64 or amd64 based on arch, otherwise linux/arm64 build of the image could be broken

yaronkoren (approved):

yaronkoren (approved):

Issue #388, opened 2024-04-19
Add relevant wikiteq branch changes to master


yaronkoren:
There have been many changes to the [wikiteq](https://github.com/CanastaWiki/Canasta/tree/wikiteq) branch since it was created about seven months ago; and though there has been an effort at synchronizing the wikiteq and master branches by myself and others, there certainly remain differences between the two - which it would be best to avoid, if possible. I went through to find changes to the wikiteq branch that it could make sense to add to the master branch in one form or another, and found the following five. They range in size from simple one-line changes (#332, #375) to a massive change to 19 files (#371), which despite its name seems to do much more than move the location of the log files - actually, much of that patch seems to itself be an attempt at synchronization, so it's probably not as big as it appears.

- #332
- #334
- #349 (note that this page contains a discussion about the merits of adding such a change to master)
- #371
- #375

It would be good to identify which of these changes it makes sense to copy over - and ideally then copy those over.

EDIT: I figured it would be good to put together, in one place, a checklist of all the tasks to be done. So, here it is - also with some improvements to the task descriptions over what is listed below:

### To do:
- [x] Fix commit SHA length (done in #394)
- [x] Update Maps extension (done in #395)
- [x] Make log files into volumes (done in #420)
- [x] Make l10n_cache directory into a volume (done in #406)
- [ ] Add `persistent directories` param to skin/extension YAML files
   - [ ] Add values for these for the SMW and Widgets extensions
- [x] Add waitelastic() function (done in #408)
- [ ] Add one or two scripts to maintenance-scripts/ directory to call the needed SMW maintenance scripts (found by calling findIncompleteTasks() - see [here](https://github.com/CanastaWiki/Canasta/blob/wikiteq/_sources/canasta/getMediawikiSettings.php#L118) and [here](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/SetupFile.php#L223)) and CirrusSearch maintenance scripts (UpdateSearchIndexConfig.php, ForceSearchIndex.php) - potentially this should instead be done by adding a `maintenance scripts` param to the skin/extension YAML files (done partially in #410 and #414)
- [x] Add `--memory-limit` flag to runJobs.php calls (done in #407)
- [x] Change `a2enmod expires` to `a2enmod expires remoteip` (done in #405)
- [x] Make sure that various directories are writeable, via the new function `make_dir_writable()` (done in #410 and #413)
[...]

PR #396, opened 2024-04-26
Enable /status and /ping for php-fpm

TechieNK:
Partial fix for #388

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,6 +66,7 @@ RUN set x; \
 	php8.1-fpm \
 	php8.1-yaml \
 	libapache2-mod-fcgid \
+	libfcgi-bin \
 	&& aptitude clean \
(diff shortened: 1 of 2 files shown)


yaronkoren:
I'm guessing it's open to the public... but does it matter?

jeffw16:
In theory, only synthetic monitors should have access to heartbeat endpoints, not the general public. It's best to find a way to prevent bad actors from being able to access this endpoint.

yaronkoren:
@vedmaka or @pastakhov - any thoughts on this? This came from your code. Is enabling /status and /ping a potential security issue, and if so, how can it be resolved?

jeffw16 (approved):

jeffw16:
After discussion at the last meeting, we decided we should hold off on this until we have some clarity about adding a firewall if needed. Rescinding my approval for now. Sorry :(

Issue #252, opened 2023-06-20
More accurately reference CDN servers


cicalese:
https://github.com/CanastaWiki/Canasta/blob/b29b7120e3d8b792e22398106ffdcda5a723bea6/_sources/canasta/CanastaDefaultSettings.php#L69

$wgCdnServersNoPurge = [];
$wgCdnServersNoPurge[] = '10.0.0.0/8';     // 10.0.0.0 – 10.255.255.255
$wgCdnServersNoPurge[] = '172.16.0.0/12';  // 172.16.0.0 – 172.31.255.255
$wgCdnServersNoPurge[] = '192.168.0.0/16'; // 192.168.0.0 – 192.168.255.255

The specified ranges are overly broad. This is preventing purges from being sent.

bawolff:
It really shouldn't matter if that setting is overly broad (other than the mild security risk if one of those servers are untrusted). What matters is that $wgCdnServers is set correctly.

cicalese:
That makes sense. Is it accurate to say that the title of the task is still correct, but the substance is that $wgCdnServers is not set, rather than $wgCdnServersNoPurge is too broad?

bawolff:
yes.

An argument could be made for not setting $wgCdnServersNoPurge to the entire private IP space like that, but its the $wgCdnServers setting that is critical.

cicalese:
Not as far as I know. I believe that @bawolff patched our instance by adding the following to CanastaDefaultSettings.php:

```
$wgCdnServers = [];
$wgCdnServers[] = "varnish:80";
```

bawolff:
I think it also involved setting $wgInternalServer to have an http protocol

cicalese:
It looks like a value of

`$wgInternalServer = http://gateway.docker.internal
`
might do the trick. Does that look correct, @bawolff ?

bawolff:
It depends on how varnish is configured. The most likely correct value would be $wgServer except http instead of https. It should be whatever caddy is sending varnish as the hist value

cicalese:
On our instance, we have caddy disabled. In this case, the IP address for gateway.docker.internal matches that of varnish. It seems that gateway.docker.internal would be correct when used with caddy as well, while specifying varnish in that case (as originally patched in our instance) would not be correct. But, somebody would need to test to be sure that works in the default Canasta case (with caddy).

cicalese:
[...]

PR #400, opened 2024-05-05
Special:Random is cached

hexmode:
## Describe the bug
**Summary**: The redirect sent by Special:Random is cached

**Description**: Visiting Special:Random initially provides a random page.   Subsequent visits to the page result in the same redirect.  Adding a param (_e.g.,_ <tt>?t=5</tt>) directs you to a new page but subsequent visits to Special:Random?t=5 end up on the same page.

## Expected behavior
A new page every time Special:Random is visited.

### DIFF
--- a/_sources/configs/default.vcl
+++ b/_sources/configs/default.vcl
@@ -54,7 +54,12 @@ sub vcl_recv {
     if (req.url ~ "/w/api.php") {
         return(pass);
     }
-    
+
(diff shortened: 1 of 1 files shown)


hexmode:
#252 is not relevant since this is just varnish caching results instead of a need to purge the cache when pages are updated (h/t @osnard).

jeffw16 (approved):

PR #405, opened 2024-06-19
changing the call a2enmod expires to a2enmod expires remoteip

naresh-kumar-babu:
Partial fix #388 .


yaronkoren:
This looks good - but why mention those two other changes in the commit message?

naresh-kumar-babu:
> This looks good - but why mention those two other changes in the commit message?

Sorry, should have mentioned it in the other PR that I closed. Removed that now.

yaronkoren (approved):

PR #407, opened 2024-06-21
addition of the --memory-limit flag (with a value of 512M) when calling runJobs.php

naresh-kumar-babu:
Partial Fix #388 .


naresh-kumar-babu:
> This looks like it contains some unrelated code changes - no?

Removed that.

yaronkoren (approved):

PR #408, opened 2024-06-21
addition of the waitelastic() function, which pauses setup until Elasticsearch starts running

naresh-kumar-babu:
Partial Fix #388 .

### DIFF
--- a/_sources/scripts/run-all.sh
+++ b/_sources/scripts/run-all.sh
@@ -205,7 +205,7 @@ check_mount_points () {
 # Wait db
 waitdatabase
 
-# Wait elastic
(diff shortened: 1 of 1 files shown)


yaronkoren:
This looks great! Could you improve the "wait elastic" comment, though, to make it a little clearer? (I'm aware that the comment that reads "wait db" is not very helpful either, but that's another story.)

naresh-kumar-babu:
> This looks great! Could you improve the "wait elastic" comment, though, to make it a little clearer? (I'm aware that the comment that reads "wait db" is not very helpful either, but that's another story.)

Modified the comment.

yaronkoren (approved):

PR #409, opened 2024-06-21
making various messages during setup more verbose

naresh-kumar-babu:
Partial Fix #388 .

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -244,4 +244,4 @@ WORKDIR $MW_HOME
 HEALTHCHECK --interval=1m --timeout=10s \
 	CMD wget -q --method=HEAD localhost/w/api.php
 
-CMD ["/run-all.sh"]
+CMD ["/run-all.sh", "-v"]


yaronkoren:
@naresh-kumar-babu - I have to apologize, because I realize now that I don't remember - and can't figure out - which parts, if any, of the original patch (#371) have to do with making setup more verbose. A search on "verbose" and "-v" does not return anything meaningful. Maybe I misread the code before? I don't know. So perhaps this whole task should just be abandoned.

Now I'm curious, though: does adding a "-v" flag to the run-all.sh call have any impact? That script doesn't seem to have such a flag, as far as I can tell.

yaronkoren:
Okay, having looked more at the code, now I think I remember what this was about: it was just the addition of a few more "echo" calls, and making some of them longer - like changing "Starting job runner..." to "Starting job runner (in 10 seconds)...". Not really that important, but it is something.

PR #410, opened 2024-06-25
Moving some code into the new files functions.sh and run-maintenance-scripts.sh

naresh-kumar-babu:
Partial Fix #388 .


yaronkoren:
This looks reasonable - looking through the code, my only comment is that run-maintenance-scripts.sh incorrectly refers to itself at the beginning and end as "run-maintenance-script.sh". (That's an error from the original Wikiteq code.)

yaronkoren (approved):

PR #411, opened 2024-06-25
Add more PHP error logging, + Apache configuration changes

naresh-kumar-babu:
Partial Fix #388 .


yaronkoren:
@pastakhov - this patch is based on some of your changes in #371, which you then explained in more detail. Does this set of changes look reasonable to you?

PR #414, opened 2024-07-04
Add get_mediawiki_cirrus_search_server function to functions.sh

naresh-kumar-babu:
Partial Fix #388 .


yaronkoren:
If I understand it correctly, this seems to really be a fix for #410 - that PR added a call to get_mediawiki_cirrus_search_server(), but did not define that function. Is that true? If so, I think the commit message should reflect that.

naresh-kumar-babu:
> If I understand it correctly, this seems to really be a fix for #410 - that PR added a call to get_mediawiki_cirrus_search_server(), but did not define that function. Is that true? If so, I think the commit message should reflect that.

Updated

yaronkoren (approved):

PR #430, opened 2024-08-23
Add Page Schemas extension

### LINKED ISSUE #348: Add Page Schemas extension
I never made a push to include Page Schemas in Canasta, even though it's my extension, in part because I remain somewhat disappointed with the way it's turned out - not quite user-friendly or flexible enough to serve as a permanent data structure maintenance tool, as far as I can tell. So I was suprised when I saw recently that, according to the "Used by" template, it's fairly popular, in use by five wiki farms and packages:

https://www.mediawiki.org/wiki/Extension:Page_Schemas

So, maybe it's more useful than I thought - and I think we might as well include it.

naresh-kumar-babu:
Closes #348.

### DIFF
--- a/_sources/configs/extensions.yaml
+++ b/_sources/configs/extensions.yaml
@@ -170,6 +170,8 @@ extensions:
   - PageForms:
       branch: master
       commit: b3be227fa9650f1a165ab1cc549100f643646714 # v. 5.7.2
+  - PageSchemas:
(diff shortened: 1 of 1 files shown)


yaronkoren:
Closing this PR - the addition of Page Schemas will now happen in a different file (contents.yaml), and probably in a different repository as well.

PR #442, opened 2024-09-26
[wikiteq] Add Extension:JsonConfig, a dependency of Extension:Graph

labster:
The dependency is needed to pass extension tests.


labster:
More than likely we will not want to merge this due to the rather extensive security issues in Extension:Graph, though JsonConfig itself is fine.

labster:
As Graph is going to be removed as of 1.43 in this branch and core Canasta, I'm going to close this.  Not that there's anything _per se_ wrong with JsonConfig, it's just unnecessary at this time.

PR #449, opened 2024-10-23
Fix 1.39.10

DanielWTQ:


DanielWTQ:
why did it create against canasta?

PR #461, opened 2024-11-25
Add PHP LDAP module (LDAPProvider not working without it)

someThor:


yaronkoren:
@someThor - thank you for this fix!

PR #465, opened 2024-12-17
Add MediaWiki bundled extensions and skins in YAML files.

naresh-kumar-babu:


yaronkoren:
This looks great! My one concern is about the Vector 2022 skin - unlike the other extensions and skins, it's not "real"; it's just a mode in the Vector skin. There's no "Vector2022" directory under skins/. There are different ways it could be possible to express this in YAML, but it seems to me that the best approach for now is just to remove the "Vector-2022" line from the file.

naresh-kumar-babu:
> This looks great! My one concern is about the Vector 2022 skin - unlike the other extensions and skins, it's not "real"; it's just a mode in the Vector skin. There's no "Vector2022" directory under skins/. There are different ways it could be possible to express this in YAML, but it seems to me that the best approach for now is just to remove the "Vector-2022" line from the file.

Removed Vector-2022

yaronkoren (approved):

PR #503, opened 2025-06-16
Version 3.0: massive overhaul, with most code moved to CanastaBase repo

yaronkoren:


.github/ISSUE_TEMPLATE/bug_report.md
jeffw16, line 1: Why was this deleted?

jeffw16 (changes requested):

naresh-kumar-babu (approved):
Looks good and everything works on testing.

jeffw16 (changes requested):

naresh-kumar-babu (approved):

.github/workflows/docker-image.yml
```diff
@@ -58,9 +52,9 @@ jobs:
           # Date
           BDATE=$(date +%Y%m%d)
 
-          # Extract MW version from Dockerfile
-          MEDIAWIKI_VERSION=$(sed -nr 's/MW_CORE_VERSION\=([0-9\.]+)/\1/p' Dockerfile | sed "s/ \\\//" | sed "s/\t//")
-          # Extract MW major version (like 1.35)
+          # We don't compute the MW version here, we just hardcode it
+          MEDIAWIKI_VERSION=1.43.0
```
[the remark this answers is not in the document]
jeffw16: How can this be sure that it will be updated each time? Would it be possible to echo it from a file like MEDIAWIKI_VERSION?

jeffw16:
@naresh-kumar-babu What’s up with the re-approval? It doesn’t override my request for changes. Please don’t try to undermine my requests for changes/clarification. I made them blocking for a reason.

yaronkoren:
I asked him to re-approve, because I didn't realize at the time that you had blocked this patch.

.github/workflows/docker-image.yml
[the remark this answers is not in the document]
jeffw16: In that case, it would be best to echo this from a MEDIAWIKI_VERSION file in the root, just like the VERSION file.

`.github` is a hidden directory and it won't be obvious to update it upon an update.

It's unfortunate there is no immediately obvious way to be absolutely sure as of now. I don't think it's worth spending additional time on making a foolproof way, but I think at the very least, hardcoding it in `docker-image.yml` could be improved upon with this MEDIAWIKI_VERSION file approach.

.github/ISSUE_TEMPLATE/support_request.md
```diff
@@ -1,43 +0,0 @@
----
```
jeffw16, line 1: Are we not doing support requests anymore? What will our users do when they make such requests anyway?

jeffw16:
I have no further concerns with the PR. Once these concerns are addressed, I'll approve this PR.

Well done team! Bravo and hats off to you!

.github/ISSUE_TEMPLATE/support_request.md
yaronkoren: Just full out a blank issue, I suppose - I think it's a reasonable interface, becuase there's a good chance that the real issue, or missing functionality, will be in CanastaBase.

.github/workflows/docker-image.yml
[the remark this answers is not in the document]
yaronkoren: That's reasonable - feel free to make the change.

.github/ISSUE_TEMPLATE/support_request.md
jeffw16: Removing the structure for people who still need the support request seems like a step backwards. Why remove it? It wasn’t doing any harm to anybody, but removing it is harmful to the volunteers who will have to waste time asking for info that could’ve already been provided had the template still existed.

.github/workflows/docker-image.yml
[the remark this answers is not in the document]
jeffw16: Would someone else be able to make the change please? I am swamped.

.github/ISSUE_TEMPLATE/support_request.md
yaronkoren: I always thought these issue templates were a little too detailed, which I suppose in some small sense counts as harmful...   but the big thing, in my opinion, is that it's directing users to the wrong place, since presumably CanastaBase is where their issues should go.

jeffw16: But it was literally removed from there too. One of them needs to get it back.

.github/workflows/docker-image.yml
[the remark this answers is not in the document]
yaronkoren: I don't know, but I'd say if you can't do it yourself, you should undo this request, or whatever the terminology is... this would be a nice feature to have, but doesn't feel like a showstopper to me.

.github/ISSUE_TEMPLATE/support_request.md
yaronkoren: Well, I did remove some of them, but that's a separate issue. (And I explained my reasoning there.) CanastaBase still has two templates, one for bugs and one for feature requests, which I think is a good number to have.

.github/workflows/docker-image.yml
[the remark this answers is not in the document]
jeffw16: Ah wait, I have an idea. Why not just add back this env var into the Canasta (not Canasta Base) Dockerfile? Easy, doesn't harm anything else, and no need to write any new code.

jeffw16 (changes requested):

.github/workflows/docker-image.yml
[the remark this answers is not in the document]
yaronkoren: Feel free to do that instead, then... it sounds like you're not completely swamped any more, which is great to hear.

.github/ISSUE_TEMPLATE/support_request.md
tosfos: I recommend a compromise, keeping the structure but with a note that says something like "Filling out all fields is optional. But if important information is missing, it is less likely that anyone will take action."

jeffw16 (approved):
Approving under duress

Issue #504, opened 2025-06-20
MW_VERSION is hardcoded


jeffw16:
## Describe the bug
**Summary**: MW_VERSION should not be hardcoded in the `.github` directory's build pipeline YAML file.

**Description**: This makes it brittle and easy to be inconsistent.

The proposal is to restore the old code:

```
MEDIAWIKI_VERSION=$(sed -nr 's/MW_CORE_VERSION\=([0-9\.]+)/\1/p' Dockerfile | sed "s/ \\\//" | sed "s/\t//")
```

and add the `MEDIAWIKI_VERSION` in the Dockerfile.

I cannot emphasize how bad this practice is to ignore simple issues and merge bad code in favor of faster velocity and fixing the issue later.

yaronkoren:
I don't know if I'd call the current setup "bad", though I agree it's not ideal. But the more I think about it, the more I think the proposed fix is not much better. There's still a chance of being inconsistent with CanastaBase - and parsing the Dockerfile to get the version number is arguably _more_ brittle, since any change to the Dockerfile formatting could break the parsing. The real solution, I would think, is to put this data in CanastaBase, so that Canasta - or any other CanastaBase-based distribution - can always be assured of getting the right value. What about having a "MEDIAWIKI_VERSION" file in CanastaBase? That way, if I understand it correctly, any child image could easily get its contents.

yaronkoren:
Fixed in #512!

PR #506, opened 2025-07-01
Version 3.0.1: upgrade to CanastaBase 1.0.1, MediaWiki 1.43.2

yaronkoren:


.github/workflows/docker-image.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           BDATE=$(date +%Y%m%d)
 
           # We don't compute the MW version here, we just hardcode it
-          MEDIAWIKI_VERSION=1.43.0
+          MEDIAWIKI_VERSION=1.43.2
```
jeffw16, line 56: Can #504 please be resolved before this is merged?

jeffw16 (changes requested):

.github/workflows/docker-image.yml
yaronkoren: That one's still under discussion, so... doubtful.

PR #512, opened 2025-07-18
Remove hardcoding of mediawiki version

naresh-kumar-babu:
Added Mediawiki version as a GitHub variable.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -52,8 +52,7 @@ jobs:
           # Date
           BDATE=$(date +%Y%m%d)
 
-          # We don't compute the MW version here, we just hardcode it
-          MEDIAWIKI_VERSION=1.43.3
+          MEDIAWIKI_VERSION=$MW_VERSION
           # Extract MW major version (like 1.43)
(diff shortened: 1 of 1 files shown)


yaronkoren:
@jeffw16 - any thoughts on this?

.github/workflows/docker-image.yml
```diff
@@ -41,6 +41,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+      -
+        name: Extract MW_VERSION from parent Docker image
+        id: extract_mw
+        run: |
+          docker build -t temp-image .
```
jeffw16, line 48: I would like to know why a temporary image has to be built instead of just reading the Dockerfile?

jeffw16 (changes requested):

.github/workflows/docker-image.yml
jeffw16: I am really concerned about the inefficiency of building a temporary Docker image just to get the version. If there’s no other easy solution, I’ll reluctantly approve this approach.

jeffw16 (approved):

yaronkoren (approved):

PR #514, opened 2025-07-24
Adds caching-proxy configuration support (MW_PROXY_SERVERS)

vedmaka:
Adds new ENV variable `MW_PROXY_SERVERS` that allows setting front-running caching proxies:

- The value can be a single server or many servers, i.e,. `MW_PROXY_SERVERS=varnish` if you're running varnish on the same stack
-  Once set, it enables `$wgUseCdn`, populates `$wgCdnServers`, enables `$wgUsePrivateIPs` to make the wiki send `PURGE` requests to the proxy servers 
- Additionally, it installs a `LinksUpdateComplete` hook that re-warms the proxy cache once the PURGE was performed to avoid cold hits


DanielButAtWork (commented):
2 minor suggestions but this has already been merged, just noting for the record

PR #534, opened 2025-10-20
Add CrawlerProtection extension

DanielButAtWork:
WIK-2113


jeffw16 (commented):
Call me biased, but I think this is a great idea.

vedmaka (approved):

PR #544, opened 2025-11-26
Remove Chameleon skin

yaronkoren:
It's not currently working, and for some reason it prevents builds from completing.

### DIFF
--- a/contents.yaml
+++ b/contents.yaml
@@ -1,1 +1,4 @@
 inherits: https://raw.githubusercontent.com/CanastaWiki/RecommendedRevisions/4f23bae9d3a6d2889caaa46648546082c94e2b64/1.43.yaml
+skins:
+- chameleon:
+    remove: true


tosfos:
Indeed. The bug is a result of a version mismatch upstream in RecommendedRevisions. [Chameleon](https://github.com/CanastaWiki/RecommendedRevisions/blob/main/1.43.yaml#L662) wants [^4.5](https://github.com/ProfessionalWiki/chameleon/blob/4.4.2/composer.json#L38). And [we're giving it Bootstrap 5.0](https://github.com/CanastaWiki/RecommendedRevisions/blob/main/1.43.yaml#L57). The result is a very sad lizard.

yaronkoren:
I don't doubt that the Bootstrap downgrade is an improvement. However, it appears to be not enough to deal with whatever strangeness the Chameleon composer call is leading to. Unless I hear otherwise, I still plan to merge this change in as soon as someone approves it. After that's done, it will of course be nice to get Chameleon working again, in one way or another.

yaronkoren:
I was wondering about that, actually - now it makes sense.

yaronkoren:
Well, 5.0.0-dev was apparently the version that existed when this MW 1.43 version of Canasta was being created. But since this version of Chameleon isn't currently working anyway, we might as well switch to the latest version...

PR #551, opened 2025-11-27
Change to custom version of Chameleon

yaronkoren:


tosfos:
Are you sure that Chameleon won't work right now?

yaronkoren:
I'm keeping this PR here until we know for sure that Chameleon works - it's certainly possible that a fork will still be needed.

PR #555, opened 2025-11-27
Restore Chameleon skin and use RecommendedRevisions with Bootstrap 5

tosfos:


yaronkoren:
Wow, this build worked! I can't believe it. So why did it fail before - was it just those large .git directories?

tosfos:
Dunno. Maybe.

yaronkoren (approved):

PR #556, opened 2025-11-30
Add a patch for WhosOnline

DanielWTQ:
Special:WhosOnline navigation building was broken due to the removal of `IndexPager::buildPrevNextNavigation()` from core, and the limit handling was problematic because it always forced the display to have the default limit instead of allowing users to configure it.


vedmaka (approved):

yaronkoren:
@DanielWTQ - sorry, I somehow missed this before. The fix to get patches working is definitely useful. And obviously this is a needed fix to WhosOnline. But did you try getting this change into the main extension code? I would think someone there would be concerned, knowing that their extension has been broken for two years...

PR #589, opened 2026-02-25
Use registry API to extract MW version without pulling image

### LINKED ISSUE #554: Consider using regclient/actions to fetch image labels without downloading image
## Background

In PR #552, we optimized the MW version extraction workflow from ~4 minutes to ~30 seconds by using labels from the CanastaBase image. However, @vedmaka suggested in [PR #552 review](https://github.com/CanastaWiki/Canasta/pull/552#discussion_r2569952911) a further optimization.

## Current Implementation

The workflow currently:
1. Builds a temporary Docker image
2. Runs it to extract `MW_CORE_VERSION` from environment variables
3. Takes ~30 seconds

## Suggested Improvement

Use [`regclient/actions`](https://github.com/regclient/actions) to fetch labels directly from the base image manifest without downloading the entire image. This would:
- Be faster (no image download)
- Use less bandwidth
- Be more efficient overall

## Considerations

- `regclient/actions` is an external third-party action (not a standard GitHub feature)
- The action has low adoption (7 stars on GitHub)
- Current solution is already well-optimized at ~30 seconds
- Would add a dependency on an external action

## Decision Needed

Should we:
1. Implement this optimization now?
2. Keep it as a future enhancement?
3. Reject it due to external dependency concerns?

## Related

- Issue #545 (original optimization request - resolved by PR #552)
- PR #552 (the optimization that was implemented)

cicalese:
## Summary

- Replace `docker pull` + `docker inspect` with lightweight `curl` calls to the ghcr.io registry API to read the `wiki.canasta.mediawiki.version` label from CanastaBase
- Avoids downloading the full ~700MB image just to read a single label, reducing the CI step from ~30s to a few seconds
- Uses only `curl` and `jq`, both pre-installed on GitHub runners — no new dependencies

## Test plan

- [x] Verify the "Extract MW_VERSION" step succeeds and prints the correct `MW_VERSION` (e.g., `1.43.6`)
- [x] Verify the "Generate tags" step still produces correct image tags containing the MW version
- [x] Confirm the step completes in seconds vs ~30s with docker pull

Resolves #554

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -49,8 +49,30 @@ jobs:
         run: |
           set -euo pipefail
           BASE_IMAGE=$(awk -F'=' '/^ARG BASE_IMAGE=/ {print $2}' Dockerfile)
-          docker pull "$BASE_IMAGE" >/dev/null
-          MW_VERSION=$(docker inspect --format '{{ index .Config.Labels "wiki.canasta.mediawiki.version" }}' "$BASE_IMAGE")
+          # Parse registry, image path, and tag from BASE_IMAGE
+          REGISTRY=$(echo "$BASE_IMAGE" | cut -d/ -f1)
+          IMAGE_PATH=$(echo "$BASE_IMAGE" | cut -d/ -f2- | cut -d: -f1)
+          TAG=$(echo "$BASE_IMAGE" | grep -o ':[^:]*$' | tr -d ':')
+          TAG=${TAG:-latest}
+          # Get anonymous bearer token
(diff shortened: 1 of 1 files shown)


tosfos:
I agree with the decision to close this. Besides the complexity and maintenance burden, this PR would have introduced other potential issues.  @vedmaka 's [suggestion](https://github.com/CanastaWiki/Canasta/issues/554#issuecomment-3589361340) was much better.

cicalese:
It was indeed Claude's summary of a much longer conversation that Claude and I had.

cicalese:
The new approach is implemented in #596.

PR #635, opened 2026-06-06
Version 3.5.9

kizule:

PR #636, opened 2026-06-08
Version 3.5.10: update base image version to 1.3.9

jeffw16:
Fixes missing change in #635


cicalese:
You're going to need to bump the Canasta version again and fix the release notes to include CanastaBase version bump in the Canasta release.

jeffw16:
@cicalese: Adjusted per your feedback

cicalese (approved):
LGTM

cicalese:
Actually, no, the release notes need to also mention the update to PageForms.

cicalese (changes requested):
Add PageForms update to release notes.

Release notes should also refer to singular "package", not packages.

yaronkoren:
Could you please change the version summary to "Update to CanastaBase 1.3.9; update Page Forms extension"? I think that's the only thing that changed.

jeffw16:
@yaronkoren: Xpdfutils was removed in 1.3.9 as well, if I understand correctly. That being said, it was a hidden change. So, do you want me to change it to "update Page Forms extension" with the understanding that Xpdfutils was also removed, but it doesn't need to be mentioned?

yaronkoren:
I believe the protocol for this version history is that it only lists changes that happened in Canasta, not CanastaBase changes (other than the changing version numbers).

yaronkoren:
Actually, on that note - the commit summary for this should really be about the 3.5.10 version.

jeffw16:
Oh right. Makes sense.

cicalese (approved):
Better.

I'm a bit conflicted about saying "update PageForms extension" since what it really updates is the version of RecommendedRevisions - but there is already a precedent for the existing language in previous release notes entries.

PR #649, opened 2026-07-01
Update contents.yaml - fixes for YAML file errors for five extensions

yaronkoren:
WhosOnline and WSOAuth were using the wrong branch (1_43 instead of master), EmbedVideo was on the wrong branch and an un-ideal commit; SyntaxHighlight_GeSHi had the wrong name (missing a "_GeSHi", though it wasn't a big problem), and Semantic MediaWiki had a "persistent directory" that was no longer in use.


cicalese (approved):
LGTM

PR #652, opened 2026-07-02
Remove dead CirrusSearch and SMW maintenance scripts

### LINKED ISSUE #651: Remove dead CirrusSearch and SMW maintenance scripts (never run in the built image)
## Summary

`cirrus-search-maintenance.sh` and `smw-maintenance.sh` are copied into the built Canasta image but are **never executed at runtime**, and haven't been in any published image. Rather than wire them up, the resolution is to **remove both scripts** (and their `Dockerfile` `COPY` lines): they are dead code, the behavior they'd provide is already handled by the documented manual workflow (CirrusSearch) or is the wiki maintainer's responsibility (SMW), and removing them simply retains the current, already-shipping behavior.

### Wrong copy destination

The `Dockerfile` copies both scripts to `/_sources/scripts/maintenance-scripts/`:

```dockerfile
COPY cirrus-search-maintenance.sh /_sources/scripts/maintenance-scripts/
COPY smw-maintenance.sh /_sources/scripts/maintenance-scripts/
```

But the base image's runtime only iterates `/maintenance-scripts/`. From CanastaBase `_sources/scripts/run-maintenance-scripts.sh`, `run_maintenance_scripts()`:

```bash
run_maintenance_scripts() {
  for maintenance_script in $(find /maintenance-scripts/ -maxdepth 1 -mindepth 1 -type f -name "*.sh" | sort); do
    ...
  done
}
```

`/_sources/scripts/maintenance-scripts/` is a CanastaBase **build-context** path: during the base build its contents are copied into the image at `/maintenance-scripts/` (`COPY _sources/scripts/maintenance-scripts/*.sh /maintenance-scripts/`). In the final Canasta image nothing ever reads `/_sources/scripts/maintenance-scripts/`, so the two scripts sit there as dead files. Verified against the published `ghcr.io/canastawiki/canasta:3.5.12` image and against Canasta `master`.

### Even if the destination were fixed, they wouldn't work as-is

- Neither script has a shebang; both are **sourced fragments** that call helpers (`run_maintenance_script_if_needed`, `get_mediawiki_variable`, etc.) defined inside CanastaBase `run-maintenance-scripts.sh`. That helper is **not** `export -f`'d, and `run_maintenance_scripts()` execs non-`mw_` scripts via `runuser -c "/maintenance-scripts/$script"` in a fresh subshell that never sources those definitions — so they'd be undefined.
- `cirrus-search-maintenance.sh` additionally calls `get_hostname_with_port`, which CanastaBase PR #193 removes from `functions.sh` — so a naive wire-up would break anyway.

In short, these have been no-ops in every published image; fixing them would be a non-trivial cross-repo change to resurrect behavior nobody currently depends on.

## Resolution: remove the scripts

Delete `cirrus-search-maintenance.sh`, `smw-maintenance.sh`, and their two `Dockerfile` `COPY` lines. No behavior change ships to users — this only removes dead files.

**CirrusSearch:** the exact three-step reindex these scripts would have run is already documented as the required manual workflow after enabling the extension (Help:Extensions and skins → "Rebuilding the search index"):

```
canasta maintenance extension CirrusSearch:UpdateSearchIndexConfig --reindexAndRemoveOk --indexIdentifier now
canasta maintenance extension CirrusSearch:ForceSearchIndex --skipLinks --indexOnSkip
canasta maintenance extension CirrusSearch:ForceSearchIndex --skipParse
```

Building the index is a mandatory, non-edge-case step when CirrusSearch is installed, so keeping it documented and explicit (rather than auto-running a potentially long reindex at container boot) is the right behavior. No doc change needed.

**SMW:** the script ran `updateEntityCollation.php` / `updateEntityCountMap.php` for SMW "incomplete setup tasks." These are obscure, only relevant after specific SMW configuration changes, and are the wiki maintainer's responsibility — an SMW maintainer making such a change knows to run them. No Canasta documentation is warranted, consistent with the earlier direction of removing SMW-specific special handling (e.g. the removed automatic `rebuildData.php` call).

cicalese:
## Summary

Removes `cirrus-search-maintenance.sh`, `smw-maintenance.sh`, and their orphaned dependencies. These scripts were copied to `/_sources/scripts/maintenance-scripts/`, but the base image runtime only executes scripts in `/maintenance-scripts/`, so neither has ever run in a published image (verified against `ghcr.io/canastawiki/canasta:3.5.12`).

Removing them retains current behavior:
- **CirrusSearch** — the index-build steps these scripts would have run are already documented as the required manual workflow on canasta.wiki (Help:Extensions and skins → "Rebuilding the search index").
- **SMW** — the `updateEntityCollation` / `updateEntityCountMap` incomplete-setup tasks are the wiki maintainer's responsibility.

## Changes

- Delete `cirrus-search-maintenance.sh` and `smw-maintenance.sh`
- Delete `getSMWSettings.php` (consumed only by the SMW script)
- Remove the `MW_MAINTENANCE_CIRRUSSEARCH_*` ENV defaults and the three `COPY` lines from the `Dockerfile` (fed only these scripts)

Fixes #651


yaronkoren:
Is "dead" really the descriptive word? They're unused and probably unnecessary, but at least some of them would provide some actual functionality if called, no?

cicalese:
Dead as in "dead code": code that is never executed.

yaronkoren (approved):

PR #682, opened 2026-07-27
Preserve empty TSV fields in the rescan report loops

### LINKED ISSUE #681: Rescan report loops still collapse empty TSV fields (unfixed-CRITICAL Component blank)
#680 fixed the tab-collapse in the jq → scan loop, but the same defect remains in the two **report-rendering** loops, which still parse the findings rows with `IFS=$'\t' read`.

It matters most in the unfixed-CRITICAL table, where the partition filter is literally `$7 == ""` — so `FIXED` is empty on **every** row of that section, and every row collapses:

```
row: … phpspreadsheet <tab> 1.29.12 <tab><tab> DataTransfer
tab-read   -> Component=[]             (FIXED swallowed "DataTransfer")
\037-read  -> Component=[DataTransfer] (FIXED=[])
```

So now that #680 makes that section render at all, its **Component column would be blank on every row** — the one column that makes those findings actionable (which extension to escalate). Easy to misread as "attribution doesn't work here," since these are the first rows that section has ever produced.

The fixable table (`$7 != ""`) is correct today only because that filter guarantees a non-empty field; it should use the same idiom rather than depend on an invariant a future filter change could quietly break — which is exactly how this bug arose.

## Fix

Apply #680's treatment to both rendering loops:

```bash
awk 'NR <= 100' "$UNFIXED_CRIT" | tr '\t' '\037' |
  while IFS=$'\037' read -r DIGEST ARCH SEVERITY VULNERABILITY PACKAGE INSTALLED FIXED EXTENSION; do
```

The `awk -F '\t'` partitions need no change — awk does not collapse empty fields (`NF=8`, `$7=[]`).

Refs #679, #680.

cicalese:
#680 fixed the tab-collapse in the jq → scan loop, but both **report-rendering** loops still parsed findings rows with `IFS=$'\t' read`, so the same defect survived where it does the most damage.

The unfixed-CRITICAL table is selected by `$7 == ""`, so `FIXED` is empty on **every** row of that section and every row collapses:

```
row: … phpspreadsheet <tab> 1.29.12 <tab><tab> DataTransfer
tab-read   -> Component=[]             (FIXED swallowed "DataTransfer")
\037-read  -> Component=[DataTransfer] (FIXED=[])
```

Now that #680 makes that section render at all, its `Component` column would have been blank on every row — the one column that says which extension to escalate, and easy to misread as "attribution doesn't work here" since these are the first rows the section has ever produced.

The fixable table is converted too. It is correct today only because its `$7 != ""` filter guarantees a non-empty field; relying on that invariant is what produced this bug in the first place.

`awk -F '\t'` partitions are left alone — awk does not collapse empty fields (`NF=8`, `$7=[]`).

## Verification

YAML parses; all `run` blocks pass `bash -n`; no `IFS=$'\t' read` remains on findings rows. Both row shapes exercised through the new idiom:

```
unfixed row  -> Fixed=[]        Component=[DataTransfer]
fixable row  -> Fixed=[4.9.11]  Component=[]
```

The real confirmation is the next scheduled scan: the headline should match the per-tag column, and the unfixed-CRITICAL rows should carry populated Components.

Fixes #681

### DIFF
--- a/.github/workflows/rescan-published-images.yml
+++ b/.github/workflows/rescan-published-images.yml
@@ -278,8 +278,8 @@ jobs:
               echo
               echo '| Digest | Platform | Severity | Vulnerability | Package | Installed | Fixed | Component |'
               echo '| --- | --- | --- | --- | --- | --- | --- | --- |'
-              awk 'NR <= 200' "$FIXABLE" |
-                while IFS=$'\t' read -r DIGEST ARCH SEVERITY VULNERABILITY PACKAGE INSTALLED FIXED EXTENSION; do
+              awk 'NR <= 200' "$FIXABLE" | tr '\t' '\037' |
+                while IFS=$'\037' read -r DIGEST ARCH SEVERITY VULNERABILITY PACKAGE INSTALLED FIXED EXTENSION; do
                   printf '| `%s` | `%s` | %s | `%s` | `%s` | `%s` | `%s` | `%s` |\n' \
                     "$DIGEST" "$ARCH" "$SEVERITY" "$VULNERABILITY" "$PACKAGE" "$INSTALLED" "$FIXED" "$EXTENSION"
                 done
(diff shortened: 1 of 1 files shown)


yaronkoren (approved):
Looks good like the previous one!

Issue #689, opened 2026-08-01
Update CrawlerProtection to 1.7.0


cicalese:
Moves the Recommended Revisions pin from `344e7b3` to `1cddaaa`.

| RR commit | Change |
| --- | --- |
| `1cddaaa` | CrawlerProtection 1.7.0 ([RR #36](https://github.com/CanastaWiki/RecommendedRevisions/pull/36)) |

CrawlerProtection 1.7.0 over the pinned 1.5.0:

- Protection is skipped in CLI and maintenance context. A maintenance script re-parsing a page that transcludes a protected special page tripped the anonymous-user check and denied itself; this is the only change that affects a wiki configuring nothing.
- `api.php` and `rest.php` entry points can now be protected, via `$wgCrawlerProtectedApiModules` and `$wgCrawlerProtectedRestPaths`. Both default to empty, so they are inert until configured.
- The IP allowlist matches on the canonical request IP, with opt-in `$wgCrawlerProtectionTrustXForwardedFor` for wikis whose proxy is not in `$wgCdnServersNoPurge`, and `$wgCrawlerProtectionTreatTempUsersAsAnon` for temporary accounts.
- New `CrawlerProtectionShouldDeny` hook for a bespoke access policy; denials are marked `noindex,nofollow`; the raw denial text moved to a message key, so it is translatable.

Every new setting defaults to off or empty, and `requires` stays MediaWiki `>= 1.39.4`.

Goes in ahead of the 3.5.20 release PR, which will then carry only the base image pin, `VERSION` and the release note.

PR #691, opened 2026-08-01
Release Canasta 3.5.20

cicalese:
Bumps the base image pin to CanastaBase 1.3.19 and `VERSION` to 3.5.20.

CanastaBase 1.3.19 serves `/server-status` to loopback over IPv6 and points the image's `HEALTHCHECK` at `127.0.0.1` numerically. The `<If>` guard had matched `REMOTE_ADDR` against `127.0.0.1` alone, so wherever `localhost` resolves to `::1` first the healthcheck reached MediaWiki instead of the status handler and failed permanently rather than retrying over IPv4.

The CrawlerProtection update named in the release note arrives via its own PR, #690 (CrawlerProtection 1.7.0). Rebased onto current master, so this PR carries only the release mechanics.

Refs #689


cicalese:
Blocked on #690. The `3.5.20` tag is emitted once, on the master push whose diff touches `VERSION`, and it is immutable — so if this merges first, `canasta:3.5.20` is built from a tree still pinning Recommended Revisions `344e7b3` (CrawlerProtection 1.5.0) and permanently disagrees with its own release note. Merge #690 first; marking ready once it lands.

yaronkoren (approved):

